PR #66, opened 2015-03-18
improved root README

drammock:
Ready for review @bambooforest.  Let me know if you have any quibbles with the wording, want to add additional papers to the final section, etc.  Otherwise should be fine to merge.


drammock:
By the way, [here is the rendered form](https://github.com/drammock/phoible/blob/better-docs/README.md)

Issue #58, opened 2014-10-12
Update language codes


bambooforest:
http://www.language-archives.org/checks/phoible.org

drammock:
Should have auto-closed when #68 was merged...

Issue #41, opened 2014-09-22
Language code update


bambooforest:
http://phoible.org/languages/daf

has been split into into Dan [dnj] and Kla-Dan [lda]

http://www-01.sil.org/iso639-3/documentation.asp?id=daf

drammock:
This was supposed to auto-close when #68 was merged.

PR #68, opened 2015-03-21
language codes check script update and references

bambooforest:
language codes check script update and finished

also some small fixes to the bib file


code/checks/language-codes.R
```diff
@@ -1,48 +1,26 @@
...
+# Check that the phoible ISO 639-3 codes are valid
 
-library(RMySQL)
+library(RCurl)
+library(dplyr)
 
-checkCodes <- function(df) {
-	codes <- df$LanguageCode
+# load data and standardize code column names
+iso639.3 <- read.csv("http://www-01.sil.org/iso639-3/iso-639-3.tab", sep="\t")
```
drammock: If you're dealing with tab-delimited files, you can just use `read.delim()` (which defaults to tab-separation) instead of `read.csv(..., sep="\t")`.

code/checks/language-codes.R
```diff
@@ -1,48 +1,26 @@
...
-	return(results)
-}
-
-# example: phoible aggregated data table - use database connection or tsv files
-con <- dbConnect(MySQL(), user="", password="", dbname="phoible", host="localhost", port = 8889)
-aggregated <- dbReadTable(con, "aggregated")
-phonemes <- dbReadTable(con, "phonemes")
-
-# assume we are in the working directory and reading from within this git repo
 aggregated <- read.csv("../../phoible-aggregated.tsv", sep="\t")
```
drammock: here you're redefining `aggregated` to be the local file... what was the point of getting the remote one then?

code/checks/language-codes.R
```diff
@@ -1,48 +1,26 @@
...
-phonemes <- read.csv("../../phoible-phonemes.tsv", sep="\t")
+head(aggregated)
+
+codes <- aggregated$LanguageCode
+c <- data.frame(codes %in% iso639.3$LanguageCode)
```
drammock: `c` is a built-in function name in R.  Don't redefine it by calling your data.frame `c`. Bad things will happen.

phoible-references.bib
```diff
@@ -2,7 +2,7 @@
 %% http://bibdesk.sourceforge.net/
 
 
-%% Created for stiv at 2014-09-13 00:58:21 +0200 
+%% Created for stiv at 2015-02-12 10:55:53 +0100
```
drammock: It's preferable to have these changes to the BibTeX file in a separate PR, since they're unrelated to the ISO code checking script.  Can achieve this by having the BibTeX changes and ISO code script changes on separate branches, instead of making all changes in your local `master` branch.

code/checks/language-codes.R
```diff
@@ -1,48 +1,26 @@
...
-
-# example: phoible aggregated data table - use database connection or tsv files
-con <- dbConnect(MySQL(), user="", password="", dbname="phoible", host="localhost", port = 8889)
-aggregated <- dbReadTable(con, "aggregated")
-phonemes <- dbReadTable(con, "phonemes")
-
-# assume we are in the working directory and reading from within this git repo
 aggregated <- read.csv("../../phoible-aggregated.tsv", sep="\t")
-phonemes <- read.csv("../../phoible-phonemes.tsv", sep="\t")
+head(aggregated)
```
drammock: interactive checking stuff like `head()` is usually best left out of scripts like this.  Useful during development but unecessary / distracting once it's published.

drammock:
I've uploaded [a spreadsheet](https://github.com/uzling/phoible-data/blob/master/bad-iso-codes.tsv) (uzling restricted access) with notes regarding how to resolve most of the remaining ISO code issues. Most of these I will fix in [my PR into this PR](https://github.com/bambooforest/phoible/pull/1). Would like to pass off the remaining few to one of the research assistants, as they require revisiting the original source documents and considering geographic or other kinds of information to determine which of the newer codes is the correct one.

drammock:
Closes #58

drammock:
Closes #41

PR #79, opened 2015-04-20
Fix iso codes

bambooforest:
This PR resolves the last 4 remaining in correct ISO codes (until next time that is). Data sources are corrected as is the references file. Tracked changes here with reasoning:

https://github.com/bambooforest/phoible/blob/fixIsoCodes/data/METADATA/iso6393-code-updates.tsv

After merge, drop me a note and I'll remove the original errors sheet and associated docs in phoible-data.


drammock:
I'm OK to merge this, but FYI I haven't been updating the bib file whenever I've made fixes to the ISO codes. It might be worth going back through the previous PR to figure out which those were, and update the bib file accordingly.

drammock:
Here is [the list of ISO codes](https://github.com/uzling/phoible-data/blob/eb7cb131746835aa24bfaccab45f9fdc65cbb879/bad-iso-codes.tsv) that were changed in #68 and in this PR.  @bambooforest can you update the bibliography accordingly?  I've never mucked with the bib files, don't wanna mess up your system :)

bambooforest:
Thanks! I suppose the ISO codes aren't that important in the bib file (if I
recall correctly not all entries have one anyway), but I'll check these.

The main thing is keeping the bib file up-to-date for the website and as
our contribution to glottolog. My not so brilliant workflow also keeps the
tertiary database's references separate (e.g. AA.bib), which get merged
into one phoible-references.bib at the end...

drammock:
I've added "discuss bibliography process" to my agenda for our next meeting.

drammock:
So now that all the bib files have been removed from the repo, is this good to merge @bambooforest ?  As a side note, what do you think about having all the PHOIBLE references in an online shared Zotero collection?

bambooforest:
Yeah this can be merged. I don't have any strong opinion regarding Zotero -- never used it. I want to make sure we keep one central phoible bib file.

PR #91, opened 2015-04-27
alleged glottal tap -> glottal stop. closes #83

### LINKED ISSUE #83: Glottal tap?
https://github.com/phoible/phoible/blob/master/data/PH/phoible_inventories.tsv#L743

Getting source material to double check.

drammock:
Called in a favor from SIL to get a scan of the original source that our grammar was citing.

> Gordon, Kent H. 1976. Phonology of Dhangar Kurux. Kathmandu: SIL and CNAS T.U.

That source uses "glottal catch" but uses glottal stop in the transcriptions, and it seems clear from the text that glottal stop is the right interpretation here. 

Ready for review/merge.

### DIFF
--- a/data/PH/phoible_inventories.tsv
+++ b/data/PH/phoible_inventories.tsv
@@ -740,7 +740,7 @@
 			"ɡ"				
 			"ɡʱ"				
 			"h"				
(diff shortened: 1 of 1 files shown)


bambooforest:
Excellent! Send me the doc and I'll add it to the squibs.

PR #92, opened 2015-04-27
fix: representation of clicks in nmn from GM

drammock:
Sorry for the extra changed lines; I used Gnumeric to do the changes and it added extra line-ending tabs. Then I used a different program to remove trailing spaces, which resulted in changes to like 80% of the lines. I'm *this close* to writing my own damn spreadsheet application, since I can't seem to get even the most basic ones to behave the way I want 'em to.

Anyway, I recommend adding `?w=1` to the end of the URL when viewing the changed files, to ignore all the whitespace changes.


bambooforest:
Looks a bit deceiving in Git, i.e. doesn't really show the tabs, but looks ok when I look in a text editor. Merging. Thanks.

PR #96, opened 2015-05-02
More agg vs gold

drammock:
WORK IN PROGRESS, do not merge yet, but feel free to comment on the diff.


bambooforest:
Looks good. I talked to Robert and he says he would just do the same thing we would do regarding Glottolog codes -- take the ISO codes and grab Glottolog codes and put them into a table with the Inventory IDs. This would be done from time to time. So far we have almost all Glottolog codes:

https://github.com/clld/phoible/tree/master/data

bambooforest:
SORRY

PR #100, opened 2015-05-08
Fix allophones

bambooforest:
noticed some notes in the allophones field, so updated those in the original data files


bambooforest:
we might consider some type of str length check on the phones and allophones field (wouldn't catch things like "rare" i guess...)

PR #104, opened 2015-05-11
fix: update glyph uniqueness test to use InventoryID instead of ISO code

drammock:
the title says it all. ready for review/merge.  On my end, this shows zero inventories with duplicate phones (yay!).


bambooforest:
What do you think about:

- keeping the InventoryID field as the ints we had before
- calling the new (temp) IDs something like PhoibleID
- creating an ISO 639-3 to GlottologID table from the current phoible data dump:

https://github.com/phoible/phoible/blob/master/phoible-aggregated.tsv

Then we can keep this mapping table up-to-date when adding new inventories, fixing ISO codes, etc. Where should it live? Under data/METADATA?

PR #108, opened 2015-06-13
Restructuring

drammock:
I think this is ready for review and merge, @bambooforest. Notes:

The inventory ID lookup is still failing (partially).  In the results folder is a text file showing each inventory in the data that did not resolve to a unique inventory ID based on the mapping table.  The lookup is based on Source and ISO code (see lines 95-96 of agg script), and if that fails, it attempts to chop off the `.pdf` file extension from the `FileNames` column and match it against `BibtexKey` in the mapping.  That works in some cases, but there are still 139 inventories that I can't resolve uniquely. So if you weren't convinced before that those inventory IDs need to get hard-written into the raw source data files, maybe now you're convinced?  If you have a clever way of doing that (and validating that it worked), I would gladly pass off that responsibility. It would have the nice side-effect of cutting the length of the agg script by about 40 lines...


bambooforest:
see issue #109 for fix todo

PR #110, opened 2015-06-17
migrated bib file to data repo

drammock:
this is sort of a test to see if PRs still work the way I expect them to now that we have the submodule thing set up. Hopefully if you merge this, the bib file will disappear from this repo (expected), but also the submodule commit that added the bib file to the data folder will magically appear in the data repo, without having to do a separate PR there.


bambooforest:
Doesn't look like it was added -- unless I missed it somewhere.

drammock:
yeah, just noticed that myself.  I'll do a second PR on the data repo and
merge it in.

bambooforest:
so with this new set up we will need to maintain pull requests and issues separately

drammock:
not sure what you mean...  PRs and issues are already separated, right?

bambooforest:
i mean PRs and issues within each sub-module / sub-repo. so should problems found in the data (say we interpreted some symbol incorrectly) be reported in phoible/data/ or in phoible/dev/ etc.? i guess the latter since we want to keep the raw-data updated and then re-aggregate?

drammock:
oh, I understand.  It does kind of make things hairy to have separate issue
trackers.  And (as we just learned) a PR into the parent module that
includes commits in the submodule does not necessarily update the
submodule.  So any PR into dev that results in changes to the data files
will now require 2 PRs: one into data and one into dev.  That is super
annoying.  Sorry I didn't catch that little wrinkle in my testing...

I'm starting to have second thoughts about this re-org.  I still like the
flexibility of being able to incorporate the new data repo as a submodule
into other projects (e.g., if we do a python or R module).  But if it makes
our workflow a pain, maybe not worth it. Thoughts?

bambooforest:
I'm not liking it so far... given that even getting it up and running has proven a pain.

PR #136, opened 2016-06-08
fix: use invID from raw sources instead of auto-gen

drammock:


drammock:
@bambooforest ready for review & squash-merge.

PR #138, opened 2016-06-09
Add new resources

bambooforest:
WIP. Still getting errors on load. Adds new data sources, too.


drammock:
when I run this branch locally, UZ runs fine but I get an error when processing the EA inventories. Looking into it now.

bambooforest:
no worries. it's probably sata formatting. i can look into it. maybe look
at the features. on way in.

PR #154, opened 2017-04-27
make trump great again

drammock:
this should resolve the trump problem.  It also includes some refactoring I did on the OrderIPA function when I was working on autobuilding features from glyphs.  After running I get

```
R> dim(final.data)
[1] 95994    46
R> dim(final.data[final.data$Trump,])
[1] 69269    46
```

ready for merge (unless you see something funny looking among the changes).


drammock:
just to be clear, this assigns `TRUE`/`FALSE` values to the `Trump` column, but does not actually remove the duplicate inventories (see the second line of the above code snippet for an easy way to do that).

aggregate-raw-data.R
```diff
@@ -21,6 +21,7 @@ data.dir <- file.path(root.dir, "raw-data")
 output.dir <- file.path(root.dir, "data")
 results.dir <- file.path(root.dir, "results")
 mapping.dir <- file.path(root.dir, "mappings")
+if (!dir.exists(results.dir))  dir.create(results.dir, mode="0755")
```
bambooforest: @drammock thanks for this -- it was driving me crazy

bambooforest:
+1 for the great PR name

PR #157, opened 2017-06-12
change trump fallback from dialect to InvID (closes #156)

### LINKED ISSUE #156: Trump duplicates
The trump procedure captures a unique list of inventories (by language code) determined by the trump hierarchy in the aggregate script when there are multiple inventories across sources, but it assigns TRUMP == True to all multiple entries (inventories) within a given source. This

final.data.trump <- final.data[final.data$Trump==TRUE,]

will still return multiple inventory descriptions belonging to the same "language".

drammock:


bambooforest:
@drammock -- i'm not sure reverting this makes any sense (because it was wrong before), but this fix that i merged is also wrong. Trump == FALSE now for all inventories that are singletons across sources, e.g. Hawaiian in SPA and UPSID

final.data[final.data$LanguageCode=="haw", ]

drammock:
Can you think is a good unit test that I could debug against?

bambooforest:
Sorry -- I spent some time trying to debug this myself, but I was having a hard time wrapping my brain around the approach. We seem to have these cases:

trump.order <- c("ph", "uz", "gm", "spa", "aa", "ea", "saphon", "upsid", "ra")

haw,spa,43
haw,spa, 352

should return: 

haw,spa,43,TRUE

Here's a crazier case (sorry for piping it):

final.data %>% filter(LanguageCode=="ell") %>% select(InventoryID, Source, Trump) %>% unique()

    InventoryID Source Trump
1           170    spa FALSE
27         2186     uz FALSE
68         2187     uz FALSE
91         2458     ea FALSE
120         318  upsid FALSE

where we should consistently always take the first or last inventory within a particular source if it's repeated (first?).

And if I'm not mistaken, these should match once trump is applied:

R> length(unique(final.data$LanguageCode))
[1] 1845
R> length(unique(final.data$LanguageCode[final.data$Trump==TRUE]))
[1] 1701

I will set myself an issue to write some unit tests asap (test this in R or do you have another preference?).

bambooforest:
@drammock -- also stumbled across this file:

https://github.com/phoible/dev/blob/master/tests/trump-reduction.R

drammock:
I think that file is obsolete now that we have inventory IDs.  I'm still trying to get to the bottom of this...

bambooforest:
Ok, let's delete this file in the next PR. I made an issue #160 to start making a list of unit tests -- this trump function has me baffled though... sorry! Thanks a lot of looking into it.

PR #159, opened 2017-06-15
Added missing Glottolog mappings.

tresoldi:
I have added the missing Glottolog mappings for four languages in Phoible: Dinka, Yaraome, Pisamira, and Puelche. I checked the inventories and the sources as far as I could, the only one which might need further investigation is the mapping for Dinka (I don't have access to the bibliography, maybe the data in Phoible refers to one of the languages/dialects of the family).

There are two languages in Phoible with no entries in Glottolog: Parkatêjê (code `qpt`) and Arara Shawãdawa (code `adc`). I will submit changes to Glottolog too, updating with the new assigned code if they are accepted.

### DIFF
--- a/gold-standard/phoible-aggregated.tsv
+++ b/gold-standard/phoible-aggregated.tsv
@@ -1396,7 +1396,7 @@ InventoryID	Source	LanguageCode	LanguageName	Glottocode	GlottologName	Trump	Lang
 1395	GM	afr	Afrikaans	afri1274	Afrikaans	1	ieur	Germanic	South Africa	Africa	4,740,000	-22.0	30.0	39	16	0	23
 1396	GM	agj	Argobba	argo1244	Argobba	1	afas	Semitic	Ethiopia	Africa	10,900	10.6596	39.756	38	30	0	8
 1397	GM	har	Harari	hara1271	Harari	1	afas	Semitic	Ethiopia	Africa	21,300	9.30485	42.1332	34	25	0	9
-1398	GM	din	Dinka			1	nsah	Nilotic	Sudan	Africa	320,000	NULL	NULL	49	20	3	26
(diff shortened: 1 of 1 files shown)


drammock:
thanks for this.  We'll have to add it elsewhere in the repository too; the `gold_standard` document is a data dump from the old SQL version of PHOIBLE so if it only goes here, it won't get propogated to future versions.  I'll leave this open until one of us gets that sorted out.

tresoldi:
I could probably do this, could you point me to the SQL source? I didn't want to fiddle in METADATA or in the actual source repositories (or what I assumed to be so) such as files in SAPHON/*

bambooforest:
Thanks a lot of these! I've made sure that the updated Glottolog codes are in our mapping table:

https://github.com/phoible/dev/blob/master/mappings/InventoryID-ISO-gcode-Bibkey-Source.tsv

which we need to add into our aggregation script, so that the output adds in the Glottolog codes (from which anyone can also merge in other data available from Glottolog, like canonical language names, latitude and longitude, etc.).

We will dump the old SQL file and add unit tests that ISO codes, Glottolog codes, etc., are up-to-date.

I'll leave this open until we add a routine to merge in the Glottolog codes from the file mentioned above.

Thanks!

drammock:
@bambooforest can we close this PR now, in light of the changes just merged in #191 ?

bambooforest:
@drammock -- yes, but they were right about "Pisamira". must be a change in the meantime. can i just hot fix the file directly for that one line since the PR is already merged? i'll remove the branch that i just pushed too.

drammock:
go for it

bambooforest:
thanks @tresoldi -- i finally got around to this (and updated other codes to Glottolog 3.3).

PR #162, opened 2017-06-23
fix: trump works properly now

drammock:
cf. discussion in #157.  Problem is that trump code would pick the most preferred `Source`, and then find the lowest `InventoryID`, and only keep the inventory if the lowest `InventoryID` happened to also be in the most preferred `Source`.  Now, the lowest `InventoryID` *within* the most preferred `Source` is kept.


drammock:
try:
```R
trumped <- subset(final.data, Trump)
# trump contains all ISO codes that we have:
length(unique(final.data$LanguageCode)) == length(unique(trumped$LanguageCode))
# trump contains the same number of InventoryIDs as ISO codes
length(unique(trumped$LanguageCode)) == length(unique(trumped$InventoryID))
# trump has exactly 1 InventoryID for each ISO code
with(trumped, all(table(LanguageCode, InventoryID) == 1))
```

Issue #143, opened 2016-06-14
Finish EA and UZ segment updates


bambooforest:
a reminder to finish the EA segment-phoneme correspondences to make them follow phoible conventions

also, update any strange segments in UZ 2016 inventories

initial input data is being added in #140

also, will need to update the bibtex mapping file and add the latest and greatest references

drammock:
In debugging the trump issue today, I noticed there are a lot of inventories within EA that have an ISO code of "0".  Any idea why?  Can you add that to the issues to fix when working on EA segments?

bambooforest:
Thanks for the hot fix. Yes, this seems to be the authors way of noting that they don't know which ISO code to assign or that there is no appropriate ISO code. We have this issue with another dataset as well (with a different way of denoting missing | unknown code), so I plan to deal with these all together.

drammock:
I suggest using "NA" rather than "0", my 2 cents.

bambooforest:
I'm going to close this issue because it's part of a larger todo which includes making sure that all segments conform to our segment conventions. See issue #160

PR #165, opened 2017-07-08
adds NAs to missing ISO codes in EA inventories, ref #143

bambooforest:
this updates that EA inventories data file that has missing ISO codes (either "0" or "-") and replaces them with "NA". see request in issues #143 (the issue isn't finished; todo).

also i have a directory that contains the entire conversion pipeline for EA (and other at least one other data source), which i think i'll put in a new repository under

https://github.com/phoible

called something like 'scripts' where each folder is its own pipeline for updating data sources via the web, e.g. if/when sources like EA update their inventories.

unless you'd prefer these little pipelines elsewhere @drammock ? it seems cleaner to me to keep them separate and run them individually when needed, and then update the raw-data in the main phoible dev / aggregation repo.

### DIFF
--- a/raw-data/EA/EA_inventories.tsv
+++ b/raw-data/EA/EA_inventories.tsv
@@ -4419,7 +4419,7 @@ InventoryID	LanguageCode	LanguageName	Segment
 2343	rue	Rusyn (Lemko)	rʲ
 2343	rue	Rusyn (Lemko)	j
 2343	rue	Rusyn (Lemko)	ɲ
-2343	rue	Rusyn (Lemko)	 ʑ
+2343	rue	Rusyn (Lemko)	ʑ
 2343	rue	Rusyn (Lemko)	ɕ
 2343	rue	Rusyn (Lemko)	dʑ
 2343	rue	Rusyn (Lemko)	tɕ
@@ -5019,7 +5019,7 @@ InventoryID	LanguageCode	LanguageName	Segment
 2356	chx	Chantyal	d
 2356	chx	Chantyal	dz
 2356	chx	Chantyal	ɡ
-2356	chx	Chantyal	m 
+2356	chx	Chantyal	m
 2356	chx	Chantyal	n
 2356	chx	Chantyal	ɲ
 2356	chx	Chantyal	ŋ
--- a/raw-data/EA/EA_inventories.tsv
+++ b/raw-data/EA/EA_inventories.tsv
@@ -391,32 +391,32 @@ InventoryID	LanguageCode	LanguageName	Segment
 2247	rjs	Rājbanshi	æ̃
 2247	rjs	Rājbanshi	ũĩ
 2247	rjs	Rājbanshi	ẽ
-2248	est	Estonian (Standard)	p
-2248	est	Estonian (Standard)	m
-2248	est	Estonian (Standard)	ʋ
-2248	est	Estonian (Standard)	f
-2248	est	Estonian (Standard)	t̪
-2248	est	Estonian (Standard)	n
-2248	est	Estonian (Standard)	s̪
-2248	est	Estonian (Standard)	l
-2248	est	Estonian (Standard)	r
-2248	est	Estonian (Standard)	tʲ
-2248	est	Estonian (Standard)	nʲ
-2248	est	Estonian (Standard)	sʲ
-2248	est	Estonian (Standard)	lʲ
-2248	est	Estonian (Standard)	ʃ
-2248	est	Estonian (Standard)	j
-2248	est	Estonian (Standard)	k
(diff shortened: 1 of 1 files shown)


raw-data/EA/EA_inventories.tsv
```diff
@@ -706,45 +706,45 @@ InventoryID	LanguageCode	LanguageName	Segment
 2255	dlg	Dolgan	ɛː
 2255	dlg	Dolgan	ɔː
 2255	dlg	Dolgan	ɑː
-2256	drh	Darkhat	b
```
drammock: Note for posterity: this language code was retired from ISO639-3 on the grounds that it is more properly considered a dialect of Halh Mongolian (`khk`)

raw-data/EA/EA_inventories.tsv
```diff
@@ -4419,7 +4419,7 @@ InventoryID	LanguageCode	LanguageName	Segment
 2343	rue	Rusyn (Lemko)	rʲ
 2343	rue	Rusyn (Lemko)	j
 2343	rue	Rusyn (Lemko)	ɲ
-2343	rue	Rusyn (Lemko)	ʑ
+2343	rue	Rusyn (Lemko)	 ʑ
```
drammock: what is this change about?

raw-data/EA/EA_inventories.tsv
```diff
@@ -5019,7 +5019,7 @@ InventoryID	LanguageCode	LanguageName	Segment
 2356	chx	Chantyal	d
 2356	chx	Chantyal	dz
 2356	chx	Chantyal	ɡ
-2356	chx	Chantyal	m
+2356	chx	Chantyal	m
```
drammock: also what is this one?

raw-data/EA/EA_inventories.tsv
```diff
@@ -8112,36 +8112,36 @@ InventoryID	LanguageCode	LanguageName	Segment
...
-2434	0	Eastern Khanty (Vakh)	yˑ
-2434	0	Eastern Khanty (Vakh)	øˑ
-2434	0	Eastern Khanty (Vakh)	ɯˑ
-2434	0	Eastern Khanty (Vakh)	ɑˑ
-2434	0	Eastern Khanty (Vakh)	oˑ
-2434	0	Eastern Khanty (Vakh)	ɪ
-2434	0	Eastern Khanty (Vakh)	ʏ
-2434	0	Eastern Khanty (Vakh)	ɯ̽
-2434	0	Eastern Khanty (Vakh)	ʊ
+2434	NA	Eastern Khanty (Vakh)	p
```
drammock: Is there a reason that we leave this as `NA` rather than assigning it to the ISO code for Khanty in general?  In other words, if we're using ISO codes as our primary classification mechanism, on what grounds are we making the judgment that ISO is wrong about this being a dialect of `kca`, but is right about, say, Darkhat being a dialect of `khk`?

drammock: (same question applies to basically all of these `NA` inventories)

bambooforest: We said NA was better than "0" or "-" which is in the raw data. They are still NA because I haven't had the time to go through and search on each and try to make an informed decision of what they could be. In general though, the Glottolog codes have a higher resolution for languages (including fine-grained dialectal information), so we may find more matches there. This brings up another issue we should discuss. It seems to me that maintaining language codes in the raw data and in the InventoryID-ISO-Glottolog is unnecessary additional work. I would suggest we keep the raw data as is (so in this case we get ISO codes from the EA compilers and leave it in the raw data), but then we implement our own language code decisions by maintaining the InventoryID-ISO-Glottolog mappings, which we add to the aggregation script. There we can add our inferences for Khanty, etc., and maintain one ISO and one Glottolog code per Inventory ID. Then it is essential that the raw data contains persistent Inventory IDs.

raw-data/EA/EA_inventories.tsv
bambooforest: There is trailing white space in the raw data, which I didn't remove previously because this gets handled by the aggregation script and the IPA segment correspondences file. I've removed the whitespace for consistency though. Good catch!

raw-data/EA/EA_inventories.tsv
drammock: agree that `NA` is better than `-` or `0`.  I'm +1 for the plan of maintaining a single mapping of InventoryID -> Glottocode -> ISOcode.  A nice feature of that would be that it could check to see whether the ISO in the raw data matched the one from the mapping doc, and issue warnings for ones that change (at least in the intial stages, that could be a useful check).

drammock:
thanks @bambooforest

Issue #167, opened 2017-07-09
Clean up bibtex file


bambooforest:
When merging all the new inventory resources, the process introduced duplicate bibtex keys (some from the original data providers themselves).

TODO:
- remove duplicates
- double-check that all inventory ID's have a bibtex (currently at least two from ER are without citations; others are URLs to online materials which need a reference added)
- make all entries valid bibtex format (when using the R bibtex package, it spits out a bunch of warnings where certain required fields are missing)
- add the relevant tests for this stuff to the tests directory

bambooforest:
PR # 168 handles removing duplicates.

drammock:
cc #210, which did some of this.  Still need to validate / implement tests.

bambooforest:
cc #240 which found some errors here.

also, a thorough clean-up would also deal with possible mismatches (as found above) between:

- filenames for squibs in GM inventories
- and filenames for squibs in GM bibtex entries

PR #168, opened 2017-07-09
Mappings

### LINKED ISSUE #146: mapping of InventoryID to Glottocode is not complete
The [mapping of InventoryID to Glottocode](https://github.com/phoible/dev/blob/ecf665c260a3bc99f1ed7b52fabb16771fb2157f/mappings/InventoryID-ISO-gcode-Bibkey-Source.tsv) is incomplete. The inventories starting at 2239 are missing.

bambooforest:
This PR updates the InventoryID - to- language code mappings (ISO and Glottolog) and updates the phoible references file.

For the index files in the mapping folder, I split this file:

InventoryID-ISO-gcode-Bibkey-Source.tsv

into 

- InventoryID-Bibtex.tsv
- InventoryID-LanguageCode-Glottocode-LanguageName-Source.tsv

because the former is in a one-to-many relationship (InventoryIDs-to-bibtex keys) and the latter is strictly InventoryID-to-unique language code. It felt like a pain to maintain the changing language codes in both places, so the former is ID-key, the latter ID-codes. I kept Source in both files because I thought it would make it easy to look up source materials, etc.

I also removed some old bib files and merged all existing and new bibtex files into the `phoible-references.bib` file under `data`.

The references file will need some work, as noted in issue #167.

Closes #146.

@drammock this PR also deletes some old files in `raw-data/METADATA`. If you think it's ok, we could also delete `raw-data/METADATA/iso6393-code-updates.tsv` (I don't see any point in keeping it since ISO codes change so frequently.) There's also a file called `inventoryids_filenames.tsv` but this feeds the phoible website, so we should keep it (but maybe move it if it ends up the only file in this directory).


mappings/InventoryID-Bibtex.tsv
```diff
@@ -0,0 +1,3832 @@
...
+2043	2043_Estradaramirez2000	saphon
+2043	2043_Estradaramirez1996	saphon
+2044	2044_Pitman1981	saphon
+2045	2045_Guillaume2008	saphon
+2046	2046_Vuillermet2006	saphon
+2047	2047_Chavarria2012	saphon
+2048	2048_Guillaume2012	saphon
+2049	2049_Ottavianoetal1965	saphon
+2050	2050_Gomez-Imbert1998	saphon
+2050
```
drammock: is that seriously the bibtex key?

drammock: URL as bibtex key?

drammock: URL as bibtex key?

drammock: URL as bibtex key?

drammock: URL as bibtex key?

drammock: URL as bibtex key?

drammock: URL as bibtex key?

drammock: URL as bibtex key?

drammock: missing newline at end of file

drammock: URL as bibtex key?

drammock: URL as bibtex key?

mappings/InventoryID-LanguageCode-Glottocode-LanguageName-Source.tsv
```diff
@@ -0,0 +1,3017 @@
...
+604	vam	vani1248	VANIMO	upsid
+605	vie	viet1252	VIETNAMESE	upsid
+606	vmb	mbab1239	MBABARAM	upsid
+607	wao	wapp1239	WAPPO	upsid
+608	wap	wapi1253	WAPISHANA	upsid
+609	wba	wara1303	WARAO	upsid
+610	wgi	nucl1620	WAHGI	upsid
+611	wic	wich1260	WICHITA	upsid
+612	wim	wikm1247	WIK-MUNKAN	upsid
+613	wnw	NA	WINTU	upsid
```
drammock: just so I'm clear: the `NA` here is because we don't have enough info in UPSID to know whether this was [Nuclear Wintu](http://glottolog.org/resource/languoid/id/nucl1651) or [Nomlaki](http://glottolog.org/resource/languoid/id/noml1242)?  If that's the case, how do we know that the right ISO code is `wnw` and not `nol`?  cf note on http://glottolog.org/resource/languoid/id/wint1259

mappings/InventoryID-LanguageCode-Glottocode-LanguageName-Source.tsv
```diff
@@ -0,0 +1,3017 @@
...
+2091	ktn	kari1311	Karitiâna	saphon
+2092	arr	karo1305	Karo	saphon
+2093	kyz	kaya1329	Kayabí	saphon
+2094	kyr	kuru1309	Kuruáya	saphon
+2095	mpu	maku1278	Makuráp	saphon
+2096	myu	mund1330	Mundurukú	saphon
+2097	yrl	nhen1239	Nheengatú	saphon
+2098	omg	omag1248	Omagua	saphon
+2099	pta	pait1247	Pai Tavytera	saphon
+2100	pak	para131
```
drammock: same question about `NA` ISO codes here as in the other comment: on what grounds do we say "this is NA" rather than, say, a dialect of [sbu](https://www.ethnologue.com/language/sbu/20)?

mappings/InventoryID-LanguageCode-Glottocode-LanguageName-Source.tsv
bambooforest: in this case, ISO split [wit] into three separate codes

http://www-01.sil.org/iso639-3/documentation.asp?id=wit

but Glottolog still refers to [wit]

http://glottolog.org/resource/languoid/id/wint1259

i updated the Glottolog code to wint1259.

mappings/InventoryID-LanguageCode-Glottocode-LanguageName-Source.tsv
bambooforest: good point. i've gone through all the codes now in mappings/InventoryID-LanguageCodes.tsv (i shortened the filename).

there are a couple of issues here.

1. before we had a few macrolanguage ISO codes, e.g.

2181	est	NA	Estonian	uz

i've updated those now so that we have the more precise ISO code [ekk] for Standard Estonian and added the Glottolog code

2. Sometimes Glottolog is more precise than ISO, for example here:

http://glottolog.org/resource/languoid/id/gunw1252

So I take the more precise Glottolog code and the parent node ISO code, e.g. Kune.

2657	gup	gund1246	Gun-Djeihmi	er
2658	gup	gune1238	Kune	er

3. Lastly, when the data is actually missing from Ethnologue, I leave it NA (but we can pick something else if you want).

2691	NA	mink1237	Minkin
2692	NA	ngum1253	Ngumbarl

In these cases it says explicitly in Glottolog that the entry is missing in ISO (happens often in extinct languages of Australia, for example).

4. When both ISO and Glottolog are NA it means there exists no code for that language yet. In  the cases of Aussie languages, they will probably get updated in Glottolog at some point and then I'll add those codes. I can't say anything regarding ISO and its update methodologies.

In a few other cases, e.g.

2264	NA	NA	Tod Tibetan	ea

I looked at some point at these, but I will revisit them in more detail. I remember there were a  few Chinese dialects.

Finally, I like our idea of keeping the codes in the original sources and dropping warnings when they don't match. I will also at some point revisit these NA NA languages.

mappings/InventoryID-LanguageCodes.tsv
```diff
@@ -3011,7 +3011,7 @@ InventoryID	LanguageCode	Glottocode	LanguageName	Source
 3010	wub	wuna1249	Wunambal	er
 3011	NA	umii1235	Umiida	er
 3012	xgu	ungg1243	Unggumi	er
-3013	NA	woro1255	Worrorra	er
+3013	wro	worr1237	Worrorra	er
 3014	NA	NA	Yawijibaya	er
 3015	wrr	ward1246	Wardaman	er
 3016	jng	yang1288	Yangman	er
```
drammock: missing newline at end of file.

drammock:
@bambooforest feel free to merge yourself once the missing newline is fixed.

Issue #98, opened 2015-05-07
Wrong language assignment


bambooforest:
you have confused two African languages under the name of Basari. One is a Gur (Voltaic) language of Northern Ghana and Togo; the other is an Atlantic language spoken in a border area between Senegal and Guinea. You have used the ISO code (bsc) and the location for the Atlantic language but have used a source describing the Ghana/Togo language, also known as Ntcham (ISO code bud) for the 
segment inventory. Thus the description and the location are seriously mismatched.

http://phoible.org/languages/bsc
http://phoible.org/languages/bud

PR #171, opened 2017-07-20
closes issue #98

bambooforest:
indeed this person was correct about the wrong language code (iirc it was Ian)

https://github.com/phoible/dev/issues/98


drammock:
yes, as I recall that was Ian.  Thanks Ian!

Issue #29, opened 2014-09-19
Wrong language


bambooforest:
the "Kxoe" language citing Christa König and Bernd Heine. 2008. A concise dictionary of Northwestern !Xun. Ruediger Koppe. > This should be ǃXun, not Kxoe/Khwe!

@bambooforest : check the language name and code

PR #172, opened 2017-07-20
Issue29

bambooforest:
addresses Christfried's comment about the wrong language code here:

https://github.com/phoible/dev/issues/29

confirmed via Glottolog references


drammock:
this PR had an extra commit in it that was already merged in 228e327.  By sheer luck, I happened to look at the two PRs in succession and in the correct order, so no rebasing was required.  You dodged a bullet!

bambooforest:
good save @drammock . i'll try to be even more careful in the future.

Issue #147, opened 2016-08-25
Map inventories to Glottolog dialects


xrotwang:
It would be useful - I guess - to be as specific as possible when mapping inventories (or the corresponding varieties) to Glottolog languoids; in particular, for cases where PHOIBLE has multiple inventories from the same source for the same Glottolog language, linking to dialects would be good.

In some cases this also seems pretty straightforward, e.g. for the [language Giryama](http://phoible.org/languages/nyf) PHOIBLE lists 6 inventories, each identified with a `LanguageName` in the source that seems to map to one of the dialects of [Giryama in Glottolog](http://glottolog.org/resource/languoid/id/giry1241).

xrotwang:
Another case is [Sebat Bet Gurage](http://phoible.org/languages/sgw), where 5 out of 6 inventories seem to correspond to [varieties on Glottolog](http://glottolog.org/resource/languoid/id/seba1251).

drammock:
agreed. can this be (semi-)automated?

bambooforest:
I started working on this issue here:

https://github.com/phoible/dev/compare/master...bambooforest:issue147

but I see that:

1. there's going to be a lot of these one-to-many mappings from ISO-to-Glottolog code
2. they aren't all so straight forward, e.g.

1465	ior	inor1238	Ennemor	gm
1466	ior	inor1238	Endegen	gm
1467	ior	inor1238	Ener	gm

http://glottolog.org/resource/languoid/id/inor1238

dialects:

Enegegny
Gyeto

the former which seems orthographically in between Ennemor and Endegen. haha.

Not that Wikipedia is better than the source dialects, but it lists:

https://en.wikipedia.org/wiki/Inor_language

"Endegegn, Enner, Gyeto, and the extinct dialect Mesmes are all sometimes considered dialects of Inor."

and the source documents describes Gumer and Ennemor, but also mentions Chaha, Soddo, and more.

Soddo is under [gru]

http://glottolog.org/resource/languoid/id/kist1241

and lists Soddo and Dobi whereas we have:

1456	gru	kist1241	Soddo	gm
1457	gru	kist1241	Goggot	gm

So regarding 1, do you have any magic method already to help out, @xrotwang ?

Regarding 2), this will take some detailed reading and consulting with Harald to match codes or update Glottolog.

xrotwang:
[...]

PR #173, opened 2017-07-27
addresses issue #147

bambooforest:
this addresses issue #147 by using the glottocode and language name match by @xrotwang to identify language names with 95% or more fuzzywuzzy match. i did a spot check on matches less than 100 and the glottocode updates look good. we could probably relax the match constraint even a bit more and maybe catch cases like:

2940	gbw	kabi1260	Butchulla	er
2941	gbw	kabi1260	Gabi-Gabi	er
2942	gbw	kabi1260	Barunggam	er

http://glottolog.org/resource/languoid/id/kabi1260

but manual investigation is probably warranted (Batyala -> Butchulla | Barunggam?)

### DIFF
--- a/mappings/InventoryID-LanguageCodes.tsv
+++ b/mappings/InventoryID-LanguageCodes.tsv
@@ -38,7 +38,7 @@ InventoryID	LanguageCode	Glottocode	LanguageName	Source
 37	tgl	taga1270	Tagalog	spa
 38	cha	cham1312	Chamorro	spa
 39	iai	iaai1238	Iai	spa
-40	khl	lusi1240	Kaliai	spa
+40	khl	kali1299	Kaliai	spa
 41	adz	adze1240	Adzera	spa
 42	mri	maor1246	Maori	spa
 43	haw	hawa1245	Hawaiian	spa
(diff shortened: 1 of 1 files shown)


drammock:
is this work-in-progress, or ready for review?

bambooforest:
@drammock i think this is ready for review. we could probably get a few more precise Glottocode matchings, but they will have to be done by hand or with the string match restraint loosened and then checked manually. for now i think this is good to go.

drammock:
OK, just so I'm clear: we're using a script to fuzzy-match our list of language names against language names in Glottolog, and accepting any match that is above 95%?  And spot-checking some of them?  Is there a further constraint on the next closest match being below 95%?  Are we only acting on languages that we already had reason to believe were matched incorrectly, or is this changing any languagename-to-glottocode pair that doesn't match what the script returned?  (sorry for all the questions; just trying to figure out how much work I need to do in reviewing this before merging).

bambooforest:
No apologies. Here's the original script from @xrotwang 👍 

https://github.com/bambooforest/glottolog-scripts/blob/master/match.py

I updated it to run on Glottolog codes in our mappings file. Basically you give it a Glottocode and language name and it looks whether there are daughter nodes (dialects) of that code with the same or nearly the same language name, and then we assign that Glottocode instead. 

So for example before we had [mina1276] for source language Besleri:

http://glottolog.org/resource/languoid/id/mina1276

and now we assign the Mina dialect code:

http://glottolog.org/resource/languoid/id/besl1239

I spot checked the ones that were not 100% string matches of the language name and they checked out, so I'm assuming that the 100% string matches are ok given the Glottocode matches.

xrotwang:
@drammock I guess the tl;dr is: The mapping **cannot** get worse above the language level, but will probably get better on the dialect level with this script.

drammock:
thanks both.  What I get now that I didn't get before is that **all** matches 95-99% were checked by @bambooforest.  I'm comfortable with that.  I just wasn't sure at first how thorough the spot-checking had been.

PR #182, opened 2018-08-19
adds australian inventories from erich round

bambooforest:
@drammock - please merge this raw data. note the InventoryID-LanguageCodes.tsv file is already up-to-date with these inventories on branch master and so is the bibtex file.


drammock:
looks like there are a lot of `NA`s in the ISO-639-3 column.  Are the inventory IDs for these languages already mapped to glottocodes?  Or is the `LanguageName` field the only identifier for those languages at present?

bambooforest:
the ID and ISO 639-3 columns match the InventoryID-LanguageCodes.tsv file (and are thus pretty much redundant, but that's how i received the data, so i kept it as is):

https://github.com/phoible/dev/blob/61e3a923561f5bd82786be7692ee63b3c4dbf212/mappings/InventoryID-LanguageCodes.tsv#L2644

it's not unusual to have Australian language varieties without ISO codes. much less so with Glottolog codes. in the end i think we move over to just Glottolog codes and rely on Glottolog for the ISO mapping.

drammock:
cool, just wanted to make sure that the mapping from `ER` inventories into Glottocodes was already in place / not accidentally missing from this PR.

PR #187, opened 2018-12-20
Refactor aggregation script

### LINKED ISSUE #158: Automatic diacritic ordering
It seems like we're messing up the order of at least one diacritic in the orderIPA function, but luckily still retaining the correct feature vectors. For example (sorry about pipes in R!):

library(dplyr)
p <- read.csv('https://raw.githubusercontent.com/phoible/dev/master/raw-data/FEATURES/phoible-segments-features.tsv',  header=T, stringsAsFactors=FALSE, sep="\t")
p %>% filter(segment=="fʷː")

vs

final.data %>% filter(Phoneme=="fːʷ") %>% select(Phoneme, 10:46) %>% unique()

but according to our conventions, the former is the preferred order. I haven't checked the other combinations, but I could create a test set for the orderIPA function.

Side note: we apply the orderIPA function to SAPHON directly, but in all other cases it gets applied in the cleanUp function. Just curious if there's a particular reason (getting the feature vectors?). And if so, should we be applying it directly on some of the other datasets?

https://github.com/phoible/dev/blob/master/aggregate-raw-data.R#L611

drammock:
Hey @bambooforest here's a stab at the aggregation refactor.  See my notes in *italics*.  Let me know which of these undone items you also want done before merging.  Note that the refactor brought to light several errors in the UZ source data (see the commit messages for the first 11 commits for details)

- [x] ~~use tidyverse for long/wide and filling missing cells~~ *not high priority for me since existing code works*
- [x] remove trump hierarchy assignment
- [x] ~~move trump hierarchy assignment to separate scripts repo~~
- [x] add Glottocodes
- [x] ~~remove ISO codes~~ *People still use/rely on these. I prefer not to remove them from our agg data until we have a mechanism in place to add them dynamically from glottolog website*
- [x] ~~rename mappings to metadata~~ *this is done separately in #190*
- [x] move `inventoryids_filenames.tsv` to metadata
- [x] delete `iso6393-code-updates.tsv`
- [x] ~~delete "gold standard"~~ keep for testing
- [x] rename final output dataframe *it now saves in the `.Rdata` file as a dataframe object called `phoible`*
- [x] dump pure CSV and not TSV  *clarify what you mean by "pure"...  It now dumps comma-separated with no field quoting and `\n` EOLs. One could argue that `\r\n` is more "pure" in that it is mandated by the IETF for CSV formats.*

closes #158


bambooforest:
@drammock this looks great to me. by pure CSV i just meant rfc 4180, and not TSV, since CSV loads by default much easier in most programs.

https://tools.ietf.org/html/rfc4180

Good catches on the weird characters! I agree with all your points in italics. I guess not in the list but the other major refactor item that we should take care of is 

[ ] add in the Australian inventories from ER (https://github.com/phoible/dev/tree/master/raw-data/ER)

Then we'll have the full list from the inventory IDs file and all the new segment types to figure out how to add.

drammock:
RFC4180 is what I was referring to with the comment about IETF mandating `\r\n` versus `\n` as the EOL character.  Usually I'm in favor of adhering to standards, but in this case I was hesitating for various reasons:

1. I *think* anyone interacting with the data from python, R, etc will probably expect `\n` and might be surprised / caught off guard by `\r\n`... probably most CSV reading functions will silently deal with it and nobody will notice, but I hate to think we might be causing annoyance or errors for a small number of researchers.
2. Anyone cloning the repo via git can get `\r\n` automatically with appropriate settings.
3. I believe Excel on Windows can open CSVs with `\n` line terminators (?)  30 minutes of internet searching suggests that Excel may struggle with CSVs that have cell-internal line breaks, but nobody seems to be complaining about reading CSVs with `\n` EOLs so I'm inferring that it just works.
4. I've seen purely pragmatic advice that it's better to stick with `\n`. For example:

> In practice, in the modern context of writing to a text file, you should always use \n (the underlying runtime will translate that if you're on a weird OS, e.g., Windows;-). The only reason to use \r is if you're writing to a character terminal (or more likely a "console window" emulating it) and want the next line you write to overwrite the last one you just wrote (https://stackoverflow.com/a/1761086/1664024)

Git's `autocrlf` function is actually the most problematic I think.  We could get around it by adding a `.gitattributes` file to the repo, specifying that the `phoible-data-by-phoneme.csv` should always be checked out with `\r\n` line endings, regardless of users' local git settings.

Does any of this change your mind? I'm about 50/50 at this point.

bambooforest:
Ok. Let’s \n then! :)

bambooforest:
@drammock - might be something wrong with the WIP merge (i can debug, but not at the moment):

library(dplyr)
phoible <- read.csv('https://raw.githubusercontent.com/phoible/dev/refactor-agg/data/phoible-by-phoneme.csv')
phoible %>% select(InventoryID, Glottocode, LanguageCode, LanguageName) %>% distinct() %>% arrange(InventoryID) %>% head()

  InventoryID Glottocode LanguageCode LanguageName
1           ˦      FALSE           uz             
2           ˧      FALSE           uz             
3         ˨ ˨      FALSE           uz             
4           1   kore1280          kor       Korean
5          10   mund1320          unr      Mundari

drammock:
hmm, I can take a look in a couple hours, thanks for the heads up

bambooforest:
Looks like these rows that contain single quotes could be messing it up:

1065,blan1242,blr,Blang; Plang; Bulang; Pulang; Pula; Kawa; K''ala; Kontoi',NA,02E6,˦,˦ ˧˥,FALSE,ph

What about dumping the CSV with quoted strings? Sounds more user friendly to me than having default csv load fail. Thanks!

bambooforest:
[ ] strip white space in the aggregation script (please :)):

library(dplyr)
load(url('https://raw.githubusercontent.com/phoible/dev/refactor-agg/data/phoible-by-phoneme.RData'))
phoible %>% filter(InventoryID %in% c(691,298,1393)) %>% distinct(InventoryID, LanguageCode, LanguageName)

bambooforest:
@drammock - are we good to go?

drammock:
yes, go for it.

PR #189, opened 2019-01-22
New branch for au data

bambooforest:


bambooforest:
This PR updates to the latest ER data, adds to the bibtex entries for ER, and updates the phoible index. Further updates to the languages codes in the index are needed in general (e.g. updating glottolog and iso codes), but I will do those wholesale in a different PR.

drammock:
@bambooforest is this ready for review & merge now?

bambooforest:
@drammock yes please.

mappings/InventoryID-Bibtex.tsv
```diff
@@ -3432,401 +3432,395 @@ InventoryID	BibtexKey	Source
...
+3011	capell_comparative_1984	er
+3012	capell_comparative_1984	er
+3013	clendon_worrorra:_2014	er
+3014	capell_comparative_1984	er
+3015	merlan_grammar_1994	er
+3016	merlan_grammar_1994	er
+3017	austin_guwamu_1980	er
+3018	nash_preliminary_1979	er
+3019	blake_dialects_2011	er
+3020	verstraete_mbarrumbathama_2018	er
```
drammock:
```suggestion
3020	verstraete_mbarrumbathama_2018	er

```

mappings/InventoryID-LanguageCodes.tsv
```diff
@@ -3009,9 +3009,13 @@ InventoryID	LanguageCode	Glottocode	LanguageName	Source
...
 3012	xgu	ungg1243	Unggumi	er
 3013	wro	worr1237	Worrorra	er
-3014	NA	NA	Yawijibaya	er
+3014	jbw	yawi1239	Yawijibaya	er
 3015	wrr	ward1246	Wardaman	er
 3016	jng	yang1288	Yangman	er
+3017	gwu	guwa1243	Guwamu	er
+3018	wbp	west2437	Ngardily	er
+3019	NA	NA	Piangil	er
+3020	lby	lamu1254	Tableland Lamalama	er
```
drammock:
```suggestion
3020	lby	lamu1254	Tableland Lamalama	er

```

mappings/InventoryID-Bibtex.tsv
```diff
@@ -3432,401 +3432,395 @@ InventoryID	BibtexKey	Source
 2625	103697	ea
 2626	testen1997ossetic	ea
 2626	phoible2283	ea
-2627	308862	ea
```
drammock: should I be worried that certain inventory IDs (like `2627` here) no longer have bib keys associated with them?

mappings/InventoryID-Bibtex.tsv
drammock: (adds missing newline to end of file)

mappings/InventoryID-LanguageCodes.tsv
drammock: (adds missing newline to end of file)

mappings/InventoryID-Bibtex.tsv
bambooforest: Yes! Good catch. I was a bit confused because the entry is in the bibtex file:

@article{308862,
	Author = {Hiram Ring},
	Bibtexkey = {ring_jowai-pnar2011},
	Date-Added = {2017-07-09 10:13:10 +0000},
	Date-Modified = {2017-07-09 10:13:10 +0000},
	Fn = {eurasia\ring_pnar2011.pdf},
	Hhtype = {phonology},
	Inlg = {English [eng]},
	Journal = {Mon-Khmer Studies: A Journal of Southeast Asian Linguistics and Languages},
	Macro_Area = {Eurasia},
	Pages = {133-175},
	Src = {hh},
	Srctrickle = {hh#phon:Ring:Pnar},
	Title = {A phonetic description and phonemic analysis of Jowai-Pnar},
	Volume = {40},
	Year = {2011}}

but somehow (or in other words I screwed up) two rows which got truncated in the `InventoryID-Bibtex.tsv` file. Sorry! Update pushed @drammock

mappings/InventoryID-Bibtex.tsv
bambooforest: @drammock isn't this optional in RFC 4180 2.2? :)

Honestly, I will be happy to put line endings at the end of all CSV files if we actually move over from TSV to CSV -- it's so much easier to deal with as the default in most programming languages and working formats.

https://tools.ietf.org/html/rfc4180

2.  Definition of the CSV Format
   2.  The last record in the file may or may not have an ending line
       break.  For example:

       aaa,bbb,ccc CRLF
       zzz,yyy,xxx

drammock: I don't object to switching these to CSVs (in a subsequent PR).  I agree that it's optional according to the standard, but git and GitHub complain if text files don't end with newlines, and it's slightly safer for anyone who wants to work with the files using typical bash tools, so I think we should standardize on including newline at EOF.

mappings/InventoryID-Bibtex.tsv
drammock: OK.  One other question.  In the old version of the file, there were several inventory IDs that had 2 entries (i.e., appeared on 2 sequential lines with 2 different bib keys).  Now all of them have only 1 bib key.  Is that intentional?  The inventory IDs in question are 2666, 2692, 2698, 2752, 2778, 2779, 2781, 2931

bambooforest: Some history here: Erich Round provided the inventories data, with his initial Glottocode assignments, a few years back:

https://github.com/bambooforest/phoible-scripts/tree/master/ER/data/raw

He had no bibtex at the time, so I went through Glottolog, by hand, and grabbed bibtex entries and named them by their Glottolog IDs for each inventory given his text references, e.g.:

https://glottolog.org/resource/reference/id/308862

Since then, Round has updated the inventories in terms of languages (e.g. some doculects got merged, others updated), but he also created his own bibtex entries for each inventory.

I decided to add in his bibtex file with ours and update our InventoryID-Bibtex.tsv from his own mapping from:

https://github.com/bambooforest/phoible-scripts/blob/master/ER/data/raw2019/inventory_biblio.txt

to:

https://github.com/bambooforest/phoible-scripts/blob/master/ER/data/raw2019/Australian_phonemes_for_PHOIBLE_20190118.tsv

because I believe he will maintain this resource more accurately than I can in the future.

For our index, I updated the bibtex entries to his 2019 inventories file:

https://github.com/bambooforest/phoible-scripts/blob/master/ER/data/formatted2019/ER_InventoryID-Bibtex.tsv

But do note that I didn't remove all the "old" bibtex entries by hand. This sounds like a lot of work that I can automate later, but right now I don't think it matter that we have more bibtex than entries.

drammock:
thanks @bambooforest

PR #191, opened 2019-01-29
updates NA Glottocodes where alternative name matches

bambooforest:
@drammock - here comes some updated Glottocodes and ISO 639-3 codes that were previously NA. @xrotwang was kind of enough to do an alternative name match for them in the latest Glottolog. some NAs remain in the index. for those NA Glottocodes, i will consult the experts and then bring up any remaining issues with the Glottolog editors. where there are NA in ISO 639-3, these occur for two reasons. 1) they are truly missing from ISO 639-3. 2) we assign a Glottocode at the most precise level, i.e. a dialect, but the appropriate ISO 639-3 code should also be inserted. in most cases i've done this by going up the tree until i hit the language level. but in some cases it's not so clear where this should be (very rare -- most cases are the former). i will implement this tree search and checks programmatically in the future.


drammock:
LGTM, thanks @bambooforest

PR #193, opened 2019-02-12
Add er data

bambooforest:
This PR simply adds the ER data to the aggregation pipeline and submits the resulting data files. @drammock please have a look if this is inline with the code style.

Also, the ER data add three segments to the unfamiliar_glyphs output:

ȴ
ȴʷ
ɺ̢

Perhaps we should decide what to do with these before merging this PR? <ȴ> seems to indicate pre-palatal lateral approximate, with a non-IPA symbol, that could be instead <ʎ̟>, I suppose, following our general conventions. 

The ER dataset in general brings up the issue of laminal places of articulation and how some Australianists divide up the space:

https://github.com/phoible/dev/tree/master/raw-data/ER

I suppose then we would remap these in our conventions to place features using the diacritics apical and laminal. What do you think?


drammock:
I'll try to get to this one tomorrow.

drammock:
This is going to take a little more thought than I originally realized.  I'm looking more closely at the README, and noticed this:

> In this dataset, laminals are labelled either as dental (with the IPA bridge diacritic) or pre-palatal (with IPA curl). Apicals are plain /t/, /n/ etc. for apical alveolars, IPA retroflex for retroflexes, and for languages with no alveolar–retroflex contrast, the single apical place is explicitly labelled as apical (with the IPA inverted bridge diacritic).

I am very pleased that this was done systematically but it raises some issues with how we use those symbols in other data sources. To first summarize how we handle coronal stops now:

- `+anterior +distributed` maps to either `t̪` or `t̻`
- `+anterior -distributed` maps to either `t` or `t̺`
- `-anterior +distributed` maps to `t̠`
- `-anterior -distributed` maps to `ʈ`
- `ȶ` is `+anterior +distributed` but also has features `+dorsal +high +front -low -back`

As far as I can tell, in the new raw ER data, as it stands now:

- a language with a laminal posterior coronal would have `ȶ` (I think `t̠` is a better choice given that textual description, unless it really is a simultaneous palatal and (post)alveolar constriction analogous to `ɕ`, in which case `ȶ` is acceptable)
- a language with a laminal anterior coronal would have `t̪`
- a language with two apical coronals would have `t` and `ʈ`
- a language with only one apical coronal would have `t̺` (this is technically allowed but not featurally distinct from plain `t`. To convey allophony between `t` and `ʈ` it would be better to either pick the "elsewhere" form as phoneme and list both forms as allophone, or encode as an UPSID-style disjunct `t|ʈ`)

The situation gets more complicated when we start dealing with central/lateral among the tap/flap/trill/approximants.  I know you want to get this in for 2.0 release but I don't think I have time for both this and the auto-feature-building changes.

drammock:
@bambooforest  I've rebased this and added a few commits, I think this is ready to go.  Please have a look... I can't add you as official reviewer since you opened the PR.

raw-data/ER/README.md
```diff
@@ -1,45 +1,121 @@
...
+of segmentation and phonemicization has been more varied. By comparing across
+sources, it is possible to discern certain recurrent patterns of analytic
+practice. The next set of topics records attempts to accommodate these issues
+in the Australian dataset. The aim has been to bring to the fore empirical
+variation, over the variability among analytic choices.
+
+## COMPLEX SEGMENTS
+
+Some sources analyse a phonotactically unusual phonetic sequence [X+Y],
+particularly ones which descend historically from *[X] or *[Y], as a complex
```
bambooforest: @drammock this appears to be italicized from here until the end.

bambooforest:
@drammock looks good except maybe some italics introduced in the markdown in the README

raw-data/ER/README.md
bambooforest: Never mind, the online version looks ok.

PR #196, opened 2019-02-12
Tsv to csv

### LINKED ISSUE #195: Convert data files from TSV to CSV
Because, you know, it's just easier for reading and writing with default program settings.

bambooforest:
Initial files in mappings directory to CSV (with extra \n ;) @drammock . I think we need to:

- [ ] decide if this is the correct CSV format (e.g. we could escape the language names that sometimes have apostrophes with double quotes) -- what is here validates via `csvclean`, but maybe we want to tailor it a bit more for processing
- [ ] update the aggregation processing script from TSV reading to CSV
- [ ] update the feature processing script (?)
- [ ] decide if we want to convert all raw data TSV files to CSV and if so to which uniformed format
- [ ] delete the TSV files

Closes #195

Note this is WIP without reviewer assigned.


drammock:
this branch may need a
```
git fetch --all --prune
git rebase upstream/master
```
...it appears to have a couple commits in it that were already merged in #194

bambooforest:
I can't tell if running those commands has fixed the issue @drammock 

But I have been saving this just for this moment:

e5e63d8d-94ee-4bf1-8e43-1120140f67b1

xrotwang:
nice! it's one of those signs with double negation, though: "no rebase forbidden".

bambooforest:
Will create a new PR instead of rebasing.

Issue #197, opened 2019-02-14
Merge indices


bambooforest:
@drammock 

I would like to merge these two files into the latter:

https://github.com/phoible/dev/blob/master/mappings/inventoryids_filenames.tsv
https://github.com/phoible/dev/blob/master/mappings/InventoryID-Bibtex.tsv

This shouldn't cause any problems and makes less clutter.

drammock:
no objections, go for it.

drammock:
closed by #199

PR #199, opened 2019-02-18
Update to csv

bambooforest:
Deals with issues #197 and converts the mappings files to CSV from TSV and updates the aggregation script to read the index as CSV. I compared the output of the aggregation script pre- and post-csv and it's the same. Good to merge.


drammock:
looks like this PR fails to actually delete https://github.com/phoible/dev/blob/master/mappings/inventoryids_filenames.tsv ?  Otherwise LGTM.

PR #200, opened 2019-02-20
adds pc from HH and DN for a few language codes

bambooforest:
does it what it says. :)


drammock:
LGTM thanks @bambooforest

PR #203, opened 2019-02-25
Issue178

### LINKED ISSUE #178: Ramaswami lacks tones
Contrastive tones reported in Ramaswami did not make it into the raw data spreadsheet. Add them from source.

bambooforest:
Fixes #178 by adding tone for the handful of languages that have it in Ramaswami source.

Also converts the raw data to CSV and updates the aggregation script to read CSV.

Removes old TSV.

After I'm finished with the other data issues @drammock , on a different branch, we'll have to run the aggregation script, which will then also update the output data to contain Ramaswami tones.


drammock:
This PR doesn't include a re-run of the agg script...  have you done it locally?  Do the tones actually make it through?  I'm wondering because there are some effectively duplicated columns (e.g., "Rising tone" vs "Low to high rising" have the same glyph, and off the top of my head I don't know if that will break aggregation, silently fail, or get handled just fine)

bambooforest:
Great call @drammock . I've now added the aggregated data. The difference is 96053-95993=60 additional rows for tones. Those numbers line up. I spot checked languages and the tone values come through correct. Hence the duplicate values don't seem to be an issue (I worked with an SE Asian specialist on the tone values and the source seems to have made some duplicate textual descriptions, e.g. "rising tone" vs "rising", which we've collapsed IPA-emically).

bambooforest:
@drammock -- will do regarding agg script and generated data. Things got a bit jumbled at some point and I was in rebase land. Dropped everything and started a new branch. My fault.

Issue #202, opened 2019-02-24
Fix nasalization diacritic on stops


bambooforest:

drammock:
closed by #205

Issue #179, opened 2018-04-11
Check sources for ṽ


bambooforest:
http://phoible.org/parameters/7B733AC6E3CC019E77B11DF92FAA7023#5/-5.084/12.248

http://phoible.org/languages/nmg
http://phoible.org/languages/umb

drammock:
closed by #205

Issue #176, opened 2017-11-26
Akha (PH) missing tones


bambooforest:
http://phoible.org/inventories/view/1162

lists no tones, the source lists 5 tonemes, page 64:

ahk_panadda1993_phon.pdf

drammock:
closed by #205

Issue #175, opened 2017-10-31
Typo in SPA allophone notes


bambooforest:
https://github.com/phoible/dev/blob/master/raw-data/SPA/SPA_Phones.tsv#L10878

(loan -> (loan)

If at some point we want to parse this additional marginal / loan information from SPA, it would be good to additionally have a spot check of this column.

drammock:
cc #150

PR #205, opened 2019-02-26
Fix remaining inventory errors

bambooforest:
This PR should handle most, if not all, of the remaining issues labelled `inventory error`, including: 

#202 
#179 
#176 
#175  
#99 (this includes removing tabbed lined as line breaks in the raw data)
#84 

And I think we can also considering closing #5, which is about not having feature vector coverage for all the clicks.


bambooforest:
@drammock sorry about those extra commits -- shot myself in the foot with a merge conflict on the Rdata object

PR #206, opened 2019-03-01
WIP about 100 bibtex keys did not match from the index to the bibtex file

bambooforest:
@xrotwang found some mismatches in the bibtex keys between our index and the .bib file. this PR fixes those.

i also have now a test script here for future updates:

https://github.com/bambooforest/phoible-scripts/blob/master/tests/tests.md

there remains one data point (from UPSID) without reference (it is like this in PHOIBLE 2014) -- cannot update.

InventoryID | BibtexKey | Source | Filename
201 | NO SOURCE GIVEN | upsid | http://web.phonetik.uni-frankfurt.de/L/L8362.html


xrotwang:
There are still a couple of citation keys with colons in them, e.g. `donaldson_ngiyambaa:_1980`. While this may be ok for BibTeX, I'd rather get rid of it, because otherwise I'd have to do it when creating the IDs for the sources in the web app.

bambooforest:
@xrotwang ok. I'll go through and update these now.

bambooforest:
@xrotwang -- colons removed. did you also want underscores removed? that would be a true pain because most of the filenames and bibtex keys have underscores in them.

xrotwang:
No, underscore is ok. There was one key with a slash  in  it, though,
removing the slash would be good.

Steven Moran <notifications@github.com> schrieb am Sa., 2. März 2019, 01:41:

bambooforest:
@xrotwang @drammock ok this should be good to go.

xrotwang:
Yes, all good.

Daniel McCloy <notifications@github.com> schrieb am So., 3. März 2019,
17:23:

PR #207, opened 2019-03-03
fixes glottocode

bambooforest:
fixes invalid glottocode


bambooforest:
@drammock i'm going to hold off on this one for the moment because i want to also update the ISO codes using them from Glottolog -- and since the aggregated data need to be rerun, i'll do it all in one PR

bambooforest:
@drammock - lots of commits here, but basically i went through the ISO codes assignment (from our previous version) and updated them to be inline with Glottolog:

https://github.com/bambooforest/phoible-scripts/blob/master/update-index/update-phoible-index.md

this involved:

- language code lookup if the glottolog code is a dialect without an iso code (we look up the tree maximally one level in the tree) -- these updates a few codes
- fixed our old codes to bring them in line with Glottolog 3.3 (most of this is from updates in ISO over the last few years); in the future we will semi-automatically update and check the ISO codes
- I added in the full canonical names from EA, which resulted also in a a glottlog code or two update
- I removed our manual assignment to "nearest language ISO code" -- there were a few like this, where I had previously assigned, say a sister language's ISO code, to the inventory. This is great for phylogenetic analyses (e.g. using the nearest neighbor in the tree), but after thinking about it:
  1. This makes future automatic assignment problematic (tests fail when we check for valid Glottolog / ISO codes, so we need to keep track of these cases...)
  2. If someone wants to do a nearest neighbor look up for an analysis, they can do it on their own

As far as I can tell, this is good to go and should be the final updates to the phoible language code index (until Glottolog gets updated). So this should be good for phoible 2.0 release.

FYI @xrotwang

PR #209, opened 2019-03-06
adds EA URIs

bambooforest:
@drammock - i got the URIs for EA inventories, so that we can link directly to them. did i screw something up here? seems like the diff is on every row. change in line ending? i don't see anything our of the ordinary.


mappings/InventoryID-Bibtex.csv
```diff
@@ -1,3844 +1,3844 @@
...
-582,Voorhoeve1965,upsid,http://web.phonetik.uni-frankfurt.de/L/L8601.html
-583,Key1968,upsid,http://web.phonetik.uni-frankfurt.de/L/L6812.html
-583,Van-Wynen1962,upsid,http://web.phonetik.uni-frankfurt.de/L/L6812.html
-584,Lynch1978,upsid,http://web.phonetik.uni-frankfurt.de/L/L2426.html
-585,Aschmann1946,upsid,http://web.phonetik.uni-frankfurt.de/L/L6713.html
-586,Bergman_etal1969,upsid,http://web.phonetik.uni-frankfurt.de/L/L4112.html
-587,Suarez1983,upsid,http://web.phonetik.uni-frankfurt.de/L/L6772.html
-588,Trumai1975,upsid,http://web.phonetik.uni-frankfurt.de/L/L6865.html
-589,Brown1973,upsid,http://web.phonetik.uni-frankfurt.de/L/L8623.html
-590,Hoijer1946,upsid,http://web.p
```
drammock: what's going on with this line?

drammock: missed this one?

drammock: missed this one?

drammock: (be sure to click through to the file...  the preview of which lines I'm commenting on is messed up.  The ones I'm asking about are inventories 2409, 2436, and 2505 --- line numbers 3217, 3246, and 3319)

bambooforest:
@drammock -- i should have mentioned that there were three NAs for filenames in EA. (now just two because i got a URI for Scots from Dmitry.) the issue here is that they are updating their database (not sure when it will be released) and they decided to "retire" these two sources (hence there no web pages):

2409,"xmf","ming1252","Mingrelian (Senaki)","ea"
2436,"xmf","ming1252","Mingrelian (Martvili)","ea"

perhaps they will just replace it with:

https://eurphon.info/languages/html?lang_id=277

but i was hesitant to use that URI twice and opted for just leaving them NA because these "filenames" will link from the phoible website to their website and two missing out of nearly 400 is ok, i guess.

At some point we could work with them to upgrade to the next version, but it seems the new data are in a nice and open SQLite database, so the whole extraction and conversion pipeline would have to be rewritten from the previous JSON formatted dump... I'm happy to put that on the longterm todo list.

Sorry, I also threw in an update to the EA readme, since he has a nice paper in LV.

PR #214, opened 2019-03-13
removes a duplicate phoneme in RA; removes space in the index

bambooforest:
@drammock - @xrotwang found some duplicates in the phonemes column. one was due to an input error on my behalf. fixed here. data regenerated.

i also removed some extra spaces in the phoible index file in the commit here -- these don't do anything to the final aggregated output since you wisely strip the language names in aggregation.

the other duplicates (only 4 or 5) are due to conventionalization. i'll email regarding those.


bambooforest:
Sorry, forgot to hit save on the raw data before regenerating the data.

Good to merge.

Issue #220, opened 2019-05-09
Update the reference to Svan


bambooforest:
https://phoible.org/inventories/view/2392

from Kevin Tuite

bambooforest:
closed by #224

PR #224, opened 2019-05-10
updates reference wrt issue #220 and request from Kevin Tuite

bambooforest:
Updates the reference by request in issue #220.


bambooforest:
Yes, he even said so in his email, which I've forwarded.

Issue #228, opened 2019-05-22
Reassign Glottolog code for Inventory ID 1984


bambooforest:
Glottolog lists the reference to Avram 2008 as [wich1264]:

https://glottolog.org/resource/reference/id/11626

"Wichí spoken in Misión La Paz-Salta-Argentina-commonly characterized as Central Pilcomayo Wichí [wich1264]"

drammock:
closed per #229

Issue #221, opened 2019-05-09
Sesotho != Sissano


bambooforest:
https://phoible.org/inventories/view/1017

is labeled Sesotho (Demuth 1992), which is correct, but it's incorrectly tagged with with the Sissano (PNG) ISO & Glottolog codes:

https://glottolog.org/resource/languoid/id/siss1243

Update to:

https://glottolog.org/resource/languoid/id/sout2807

sout2807
sot

https://github.com/phoible/dev/blob/master/mappings/InventoryID-LanguageCodes.csv#L1018

drammock:
closed by #223

PR #229, opened 2019-05-22
reassigns glottocode for issue #228

bambooforest:
I updated a Glottocode to make the reference in line with the Glottolog. @drammock do you want me to regenerate the dumps as well before merging?


bambooforest:
i regenerated data @drammock . includes regenerated data for #221 (will also regen each issue moving forward).

PR #235, opened 2019-07-24
Issue232

bambooforest:


bambooforest:
@drammock - note that we have changes like this in UPSID_IPA correspondences:

85	d̠n̠	d̠ⁿ̠		d_n	nasally-released voiced palato-alveolar plosive
92	ɖɳ	ɖⁿ		d.n	nasally-released voiced retroflex plosive

where the UPSID description doesn't seem to be fit our convention (pre-stopped nasal). this seems OK because:

1. it's one language from UPSID (https://phoible.org/inventories/view/201)

2. this data point is the only one there that lacks a reference

2. it puts the PI in line with recent descriptions of the language (https://phoible.org/inventories/view/2158)

bambooforest:
Will do a different PR that doesn't change UPSID_IPA correspondences.

PR #237, opened 2019-07-25
Conventionalizes pre-nasalization in EA

### LINKED ISSUE #232: Prenasalization vs nasal release
@drammock - the immediate issue is what to do regarding the UPSID_IPA correspondences:

74	nd	ⁿd		nd	prenasalized voiced alveolar plosive	
75	dⁿ	dⁿ		dn	nasally-released voiced alveolar plosive

https://github.com/phoible/dev/blob/master/raw-data/UPSID/UPSID_IPA_correspondences.tsv#L75-L76

if I change /dⁿ/ -> /nd/, etc., we collapse a distinction in UPSID that apparently only exists in ARRERNTE (as discussed over email and summarized for this issue below). Given that ARRERNTE doesn't have a citation in UPSID and this would bring it more in-line with with phoible conventions, I'm OK with it.

Note also open issue #61 .

Regarding prenasalization vs nasal release in phoible, there are two issues.

```
load("phoible.RData")
library(dplyr)
phoible %>% filter(grepl("ⁿ", Phoneme)) %>% select(LanguageName, Phoneme, InventoryID)

          LanguageName Phoneme InventoryID
1             ARRERNTE      bⁿ         201
2             ARRERNTE      dⁿ         201
3             ARRERNTE      d̠ⁿ         201
4             ARRERNTE      d̪ⁿ         201
5             ARRERNTE      ɖⁿ         201
6             ARRERNTE      ɡⁿ         201
7    Caodeng rGyalrong      ⁿb        2257
8    Caodeng rGyalrong      ⁿd        2257
9    Caodeng rGyalrong     ⁿdz        2257
10   Caodeng rGyalrong     ⁿd̠ʒ        2257
11   Caodeng rGyalrong     ⁿɖʐ        2257
12   Caodeng rGyalrong      ⁿɡ        2257
13   Caodeng rGyalrong      ⁿɢ        2257
14   Caodeng rGyalrong      ⁿɟ        2257
...
```

1. ARRERNTE

- proposed change to mb, nt, etc.
- because the phonological analysis emphasizes that these are prenasalized stops, not sequences/clusters, which is why they avoid /mb/.  But in phoible all the contrastive segments are separated, so just the fact that we have an entry for /mb/ makes it clear that it's not a sequence/cluster.

2. EURPhon inventories

- propose change to mb, nt, etc.
- because EURPhon developers use nasal release superscript for prenasalization

bambooforest:
Closes #232


raw-data/EA/EA_IPA_correspondences.tsv
```diff
@@ -1,1487 +1,1487 @@
...
+n̥̪	n̪̥	NA
+nː	nː	NA
+n̥ː	n̥ː	NA
+ⁿb	mb	conventionalized
+ⁿc	ɲc	conventionalized
+ⁿcʰ	ɲcʰ	conventionalized
+ⁿd	nd	conventionalized
+ⁿd̪	n̪d̪	conventionalized
+ⁿdz	ndz	conventionalized
+ⁿd͇z͇	n͇d͇z͇	conventionalized
```
drammock: This is a case where the diacritic should not spread to the nasal.  It's "non-sibilant" (hence a *manner* diacritic) and I think we generally only want to match *place* diacritics.

raw-data/EA/EA_IPA_correspondences.tsv
```diff
@@ -1,1487 +1,1487 @@
...
+ⁿpʰ	mpʰ	conventionalized
+ⁿq	ɴq	conventionalized
+ⁿqʰ	ɴqʰ	conventionalized
+ⁿsʰ	nsʰ	conventionalized
+ⁿt	nt	conventionalized
+ⁿtɕʰ	ntɕʰ	conventionalized
+ⁿtʰ	ntʰ	conventionalized
+ⁿts	nts	conventionalized
+ⁿtsʰ	ntsʰ	new segment type added
+ⁿt͇s͇ʰ	n͇t͇s͇ʰ	conventionalized
```
drammock: ditto here, no diacritic spreading to n

bambooforest:
@drammock - keen eye. thanks.

PR #239, opened 2019-07-25
Issue238

### LINKED ISSUE #238: Incorrect bibtex key
Here I've listed two of the same bibtex keys (1496_beledugu_bam) for two different sources:

https://github.com/phoible/dev/blob/master/mappings/InventoryID-Bibtex.csv#L2183-L2184

TODO: add Chris Green's (bam_green2010.pdf) dissertation to the bibtex file and update the InventoryID-Bibtex.csv mapping file.

Another reason to write tests as per #167.

bambooforest:
Closes #238

### DIFF
--- a/data/phoible-references.bib
+++ b/data/phoible-references.bib
@@ -1,3 +1,21 @@
+%% This BibTeX bibliography file was created using BibDesk.
+%% http://bibdesk.sourceforge.net/
+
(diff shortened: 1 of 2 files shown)


bambooforest:
@drammock - didn't regen data on this one, since we don't dump refs in the phoible csv file.

PR #242, opened 2019-07-29
closes #240

### LINKED ISSUE #240: Misalignment in bibtex keys and filenames
Somewhere here there's a one-off alignment problem:

https://github.com/phoible/dev/blob/master/mappings/InventoryID-Bibtex.csv#L1916-L2018

Maybe duplicate "khy_Coupez1977.pdf	" at the end.

Fix.

bambooforest:
Fixes the alignment of the bibtex keys and filenames (thanks @EladEisen !).

Also updates some of the filenames for squibs in the bibtex file. More of these are probably wrong, but cc #167 - but it's not mission critical, I think, to have filenames (note IDs are OK!) in the bibtex .bib file and the bibtex mappings file (although at some point it would be nice to either align them or remove them from the bibtex file).

Also regen'd the data, but there's no difference.


mappings/InventoryID-Bibtex.csv
```diff
@@ -3841,4 +3838,4 @@ InventoryID,BibtexKey,Source,Filename,URI
 3017,austin_guwamu_1980,er,NA,NA
 3018,nash_preliminary_1979,er,NA,NA
 3019,blake_dialects_2011,er,NA,NA
-3020,verstraete_mbarrumbathama_2018,er,NA,NA
+3020,verstraete_mbarrumbathama_2018,er,NA,NA
```
drammock: missing newline at end of file.

PR #247, opened 2019-10-25
sort before saving

drammock:
This should (hopefully) eliminate spurious git diffs caused by different row order, when regenerating aggregate data in each pull request.


scripts/add-features.R
```diff
@@ -85,7 +85,7 @@ make_feat_vec_from_mat <- function(feat_mat) {
     ## collapse just the base glyphs into a string
     bases <- paste(feat_mat[feat_mat$GlyphType %in% "B", "segment"],
                    collapse="")
-
```
drammock: this appears to be a tab on an empty line, auto-deleted by my text editor

scripts/aggregate-raw-data.R
```diff
@@ -248,6 +248,10 @@ all_data$GlyphID <- get_codepoints(all_data$Phoneme)
 ## CONVERT INVENTORY ID TO INTEGER
 all_data$InventoryID <- as.integer(all_data$InventoryID)
 
+## SORT BEFORE SAVING TO ELIMINATE SPURIOUS GIT DIFFS
+sort_order <- with(all_data, order(InventoryID, SegmentClass, GlyphID))
```
bambooforest: @drammock - so this sorts by InventoryID, then consonants < tones < vowels, then by GlyphID. but the GlyphIDs are treated as strings and not decimal ordering, I think:

>>> int("006A", 16)
106
>>> int("0068", 16)
104
>>> int("0077", 16)
119
>>> int("0061", 16)
97

If it's consistent throughout, I'm fine with it as is.

bambooforest:
@drammock -- sorry didn't even see this until I made PR #249. Note once we merge this, I'll go back and regen the data for that PR.

I ran this PR on my fork and the CSV file doesn't change, but the Rdata object does -- perhaps due to serialization.

See comment on the ordering.

My only other concern about the Rdata object is the rownames, which are now ascending, but not by increments of one, e.g.:

      InventoryID Glottocode ISO6393
1               1   kore1280     kor
3282            1   kore1280     kor
6976            1   kore1280     kor
7265            1   kore1280     kor
7760            1   kore1280     kor
9328            1   kore1280     kor
11681           1   kore1280     kor
13988           1   kore1280     kor
16540           1   kore1280     kor
17738           1   kore1280     kor
18256           1   kore1280     kor

Should we reset the rownames before dumping to Rdata?

scripts/aggregate-raw-data.R
drammock: GlyphID has to be treated as a string, because of cases like `006B+02B0`. If you want to sort by something else instead (phoneme?) I can change it.  As far as storing in string vs decimal integer or hexadecimal integer, I think string is best because it straightforwardly reflects unicode codepoint values that people are likely to encounter elsewhere.  Converting the representation of `kʰ` from `006B+02B0` to something like `107,688` seems like a move away from usability in my opinion.

drammock:
> Should we reset the rownames before dumping to Rdata?

Yeah, we should do that.

drammock:
OK @bambooforest if you're cool with sorting by Phoneme instead of GlyphID, this should be ready to go.  If not, LMK how you think sort should be handled.

drammock:
(note in case we need it: R functions `strtoi()` and `as.hexmode()`)

bambooforest:
@drammock - thanks! I'm ok with just sorting on phonemes for diff's sake. I think ideally (or perhaps just linguistically we could sort on "the IPA chart order", e.g. p, b, t, k, ... ), but there's no straightforward way (off the top of my head) to do this. I do like your consonants < tones | vowels though.

drammock:
We could approximate IPA order by sorting on feature columns. But to me that seems like overkill; it would require a lot of nested sorting. I'm content with just getting it to be deterministic, which it should be now.

-------- Original Message --------

Issue #253, opened 2020-04-23
"!" found its way into SpecificDilaect column


bambooforest:
https://github.com/phoible/dev/blob/master/raw-data/UZ/UZ_inventories.tsv#L956-L969

drammock:
fixed by #254

Issue #160, opened 2017-06-21
Write unit tests


bambooforest:

bambooforest:
some units tests include:

- known number of unique language codes
- range of segment counts (11-161)
- NAs in various columns, including the feature columns, language codes, etc. (i guess any column that can't contain a null)
- duplicate feature vectors (note already in aggregation script)
- missing feature vectors (also already in aggregation script)
- column type checks (we might want to cast certain columns to numeric, character, etc.)

bambooforest:
http://kbroman.org/pkg_primer/pages/tests.html

for unit tests. it'd be nice to integrate travis as well:

https://travis-ci.org/

bambooforest:
@drammock do you have a preferred testing framework for R? i see a few different options, like

https://github.com/yihui/testit
https://github.com/hadley/testthat

a quick-and-dirty approach would be to write tests that run on the output of the aggregation script (language code checks, segment convention validation, that all bibtex keys match InventoryIDs, etc.)

a more in-depth approach would be to integrate the tests into a phoible R package. i guess the question for me is whether we move forward with an R package by writing the tests, thus skipping the quick-and-dirty in part (tests of course still need to be written :)

what do you think?

drammock:
I haven't done proper test-driven development in R before, only ad-hoc testing scripts.  I like the simplicity of `testit` (the README was enough that I already know how to use it, whereas `testthat` says "go read this whole chapter (and probably, all the chapters that came before it)."  Honestly I've been burying my head in the sand about writing tests because I secretly want to switch to python (for which I use `nosetests`).  Might be time for a skype meeting so you can talk me down from triggering another overhaul :grimacing:

bambooforest:
We could also ideally include continuous integration

bambooforest:
Note some tests I've written here should be eventually incorporated into dev:

https://github.com/bambooforest/phoible-scripts/blob/master/tests/tests.md

bambooforest:
@bambooforest -- create the list of tests that we need (as per our discussion June 2, 2020)

bambooforest:
tests failing here:

- inventory 2478 is the one failing the phoneme uniqueness test.
- invalid ISOs: "daf" "duj" "wit" "kxl" "krm" "kxu" NA "drh" "dit" "mwd" "qgu" "bjd" "ggr" "yiy"
- invalid glottocodes: NA is the only one
- inventory 695 has one row where the language name is "dida" instead of "Dida" (capitalization)...

PR #254, opened 2020-04-23
Issue253

bambooforest:
- removes "!" in the SpecificDialect column, as per #253
- regens the data
- adds a test for making sure the number of unique inventory ids == unique inventories + the metadata columns

means we should also revisit #160 some day and integrate data tests reasonably (e.g. CI tests)


tests/tests.R
```diff
@@ -0,0 +1,11 @@
...
+
+library(dplyr)
+library(testthat)
+
+load('../data/phoible.RData')
+
+# Metadata categories should be unique within inventories
+ids <- phoible %>% select(InventoryID) %>% distinct()
+metadata <- phoible %>% select(InventoryID, Glottocode, ISO6393, LanguageName, SpecificDialect, Source) %>% distinct()
+expect_equal(nrow(ids), nrow(metadata))
```
drammock, line 11: missing newline at EOF

bambooforest: @drammock - i hate the new line. :)

drammock: You should set your text editor to put it in automatically when saving the file. :)

PR #260, opened 2020-05-10
closes https://github.com/cldf-datasets/phoible/issues/7

bambooforest:
Correct Glottocode for Mbe:

https://github.com/cldf-datasets/phoible/issues/7

I also updated the ISO codes and squib names and link to the reference.

Regens the data.


raw-data/GM/gm-afr-inventories.tsv
```diff
@@ -17722,7 +17722,7 @@ InventoryID	LanguageCode	LanguageName	SpecificDialect	Phoneme	Notes	Allophones	A
 	mif	Mofu, South	Gudur	aː				
 	mif	Mofu, South	Gudur	e				
 	mif	Mofu, South	Gudur	eː				
-1599	mtk	Mbe		i				mtk.pdf
+1599	mfo	Mbe		i				mfo_Pohlig1981.pdf
 	mtk	Mbe		e
```
drammock, line 17726: only changed the ISO for the first row... need to change the other rows too?

bambooforest: @drammock -- good catch. thanks! we should be consistent, but tbh we don't need this column at all (but updating the raw data and code ... ugh).

PR #261, opened 2020-05-12
closes https://github.com/cldf-datasets/phoible/issues/9

bambooforest:

PR #262, opened 2020-05-12
closes with issues 8, 10, 11 here: 

bambooforest:
https://github.com/cldf-datasets/phoible/issues

sorry @drammock - these could have been together with the previous PR


raw-data/PH/phoible_inventories.tsv
```diff
@@ -4434,7 +4434,7 @@
 				"eː"				
 				"aː"				
 
-962	"igo"	"Ngomba"	"Not finished, became too confusing. To finish later"	"i"		"[ɨ]"	"In dipthongs where the second vowel is ə"	"igo_ngouagna1988.pdf"
+962	"jgo"	"Ngomba"	"Not finished, became too confusing. To finish later"	"i"		"[ɨ]"	"In dipthongs where the second vowel is ə"	"igo_ngouagna1988.pdf"
```
drammock, line 4437: the comment on this line is a bit unsettling...  is that cruft of is this entry really unfinished?

bambooforest:
@drammock -- so we we have the same issue here wrt to the Rdata file.

drammock:
> @drammock -- so we we have the same issue here wrt to the Rdata file.

yep. I've now merged #261 so you can rebase this branch against upstream/master now and it should resolve the merge conflict.  After the rebase you should re-run the aggregation scripts and re-add the RData file though; otherwise it will be out-of-sync with the CSV version of the data.

raw-data/PH/phoible_inventories.tsv
bambooforest: Will address in #263

PR #273, opened 2020-06-03
Ditch binary

drammock:
The diff in `phoible.csv` is due to the fact that `FALSE` and `TRUE` in the `marginal` column were getting quoted in past versions of the csv, but aren't getting quoted now.  Possibly a mac-vs-linux difference, or a difference in default settings in `options()` and/or `.Rprofile`.  Hopefully, though, it's a one-time thing.


bambooforest:
@drammock -- I'm going to merge this, but one thing that I was thinking it that we could keep it, but just include in the .gitignore a line for `RData` files. Then at least they would be created locally, but ignored in adds/commits.

PR #274, opened 2020-06-04
Issue271

bambooforest:


bambooforest:
@drammock -- no changes to regenerating the data here because this only changes the mapping file's links to the references. this seems to be past bug in assigning references to the AA source, e.g. when there were multiple doculects for the same language, reference assignment was mixed.

drammock:
changes look reasonable, but I'm basically taking it on faith that you know what you're doing here, since you're the bibliography master.

bambooforest:
I'll add that title to my CV.

PR #275, opened 2020-06-07
WIP: update / modernize some tests

drammock:
This is in preparation for setting up CIs to run the tests.

TODO:

- [x] update all tests in the `tests` directory to use `testthat` (or move/delete them... e.g. the gold-standard one)
- [x] rename test files to start with `test-`
- [x] add `.github/workflows/main.yml`

@bambooforest note that neither of the tests I have updated so far are actually passing... there is 1 inventory with non-unique phonemes, and several invalid ISO codes (though all glottocodes are valid except for `NA`)


drammock:
@bambooforest I'm going to go ahead and merge this to see if having the yaml file in master is enough to trigger the workflow (so far it's not triggering)

drammock:
looks like it's working: https://github.com/phoible/dev/runs/775439974?check_suite_focus=true

PR #281, opened 2020-06-22
Issue265

bambooforest:
See email.


drammock:
These changes go in the wrong direction.  We should not use creaky-voicing diacritic on voiceless base glyphs.

If I look only at Beeler 1970 (the basis for inventory 1239) I would say we should in 1239 change ejective diacritic to 02C0 (ˀ) because "glottalized" is never explained further.  However, Wash 2001 (the basis for inventory 862) has a brief comment that "glottalization is so frequent in Barbareño that it gives the language a 'staccato effect'."  This to me is a fairly good indication that what is meant by "glottalization" by Wash is in fact "ejective". Wash cites Beeler 1976 regarding the "staccato effect" observation, so we can infer that Beeler probably also meant "ejective" when saying "glottalized" (even though the "staccato" comment isn't made in that particular paper).  On that basis I would leave ejectives in 1239 alone, and change creaky diacritic to ejective diacritic in 862.

I would also be inclined to add a note to the ejective phonemes in 1239 saying "subsequent literature (e.g. Wash 2001) suggests that glottalized==ejective for Beeler", and a similar note for ejective phonemes in 862 saying "glottalized==ejective based on comment about staccato sound of language" (or something similar).

bambooforest:
@drammock -- updated as discussed. ready for review.

raw-data/PH/phoible_inventories.tsv
```diff
@@ -481,20 +481,20 @@
...
+				"ˀn"				
 				"l"		"[l̥]"	"""Syllable-finally and word-finally"""	
-				"l̰"				
+				"ˀl"				
 				"w"		"[w̥]"	"""Syllable-finally and word-finally"""	
 				"j"		"[j̥]"	"""Syllable-finally and word-finally"""	
-				"w̰"				
-				"j̰"				
+				"ˀw"				
+				"ˀj"
```
drammock, line 497: These aren't described as "pre-glottalized" so the diacritic should be on the right of the base glyph.

raw-data/PH/phoible_inventories.tsv
```diff
@@ -481,20 +481,20 @@
 				"<sʰ>"	"Loan"			
 				"ts"				
 				"t̠ʃ"				
-				"ts̰"				
-				"t̠ʃ̰"				
+				"tsʼ"				
+				"t̠ʃʼ"
```
drammock, line 485: these rows should get the comment about "staccato" too

raw-data/PH/phoible_inventories.tsv
```diff
@@ -15409,27 +15409,27 @@
...
-				"w̰"				
+				"ˀl"				
+				"ˀj"				
+				"ˀw"				
 				"m"				
 				"n"				
-				"m̰"				
-				"n̰"				
+				"ˀm"				
+				"ˀn"
```
drammock, line 15432: These aren't described as "pre-glottalized" so the diacritic should be on the right of the base glyph.

raw-data/PH/phoible_inventories.tsv
```diff
@@ -15409,27 +15409,27 @@
 				"ʃ"				
 				"x"				
 				"h"				
-				"sʼ"				
-				"ʃʼ"				
-				"xʼ"				
+				"sʼ"	"""Subsequent literature (by Wash 2001) suggests that glottalized means ejective for Beeler based on Beeler 1976 mentioning a 'staccato effect'."""			
+				"ʃʼ"	"""Subsequent literature (by Wash 2001) suggests that glottalized means ejective for Beeler based on Beeler 1976 mentioning a 'staccato effect'."""			
+				"xʼ"	"""Subsequent literature (by Wash 2001) suggests that glottalized means ejective for Beeler based on Beeler 1976 mentioning a 'staccato effect'."""
```
drammock, line 15414: Having looked back at the sources, I think we ought to consider x and xʼ as uvular, not velar.  They are aligned in the same column as q, not k.  This is true in inventory 862 also: x is in the q column, not the k column.

bambooforest:
@drammock updates pushed.

PR #286, opened 2020-07-04
Add Abun inventory by Brynna

bnk7:
The last three commits are the only ones that make any changes. The others are there because my repository got out of sync, and they were included in the previous pull request.


raw-data/UW/uw_inventories.tsv
```diff
@@ -53,3 +53,43 @@ roseslabrada2015mako.pdf		Mako		i		i
 						n	"/l/ can also be variably nasalized when followed by nasal vowels"	
 				w		w		
 				ˀw		ˀw		
+
+berry1999abun.pdf	abun1252	Abun		p		p
```
drammock, line 57: Given what the author says about where the data were collected (p3 section 1.4, and p8-9 section 1.7) I think it makes most sense to treat this as a description of Abun Ye dialect. Please add that to the "specific dialect" column, and with that in mind, you can remove notes and allphones that refer to other dialects.

Note that in principle we could consider creating four separate inventory entries (one for each dialect), all based on this one document, if the description of each was sufficiently thorough.  But in this case they explicitly say where their data came from (both villages are within the Abun Ye territory)

raw-data/UW/uw_inventories.tsv
```diff
@@ -53,3 +53,43 @@ roseslabrada2015mako.pdf		Mako		i		i
...
+				ʃ		ʃ		
+						t̠ʃ	"A voiceless palatal affricate [tš] occurs in the Abun Ji dialect on occasion in free variation with the palatal fricative [š]"	
+						s	"/š/... neutralises to /s/ syllable-finally"	
+				m		m		
+				n		n		
+				ɲ		ɲ		
+						n̠d̠ʒ	"Neither does the /nj/ phoneme occur in the Abun Tat dialect. Rather it is an allophone of /ñ/"	
+						n	"/ñ/... neutralises to /n/ word-finally"	
+				w		w		
+				r		r
```
drammock, line 79: The table on page 10 lists this as "glide" which suggests `ɹ` not `r` (a trill).

raw-data/UW/uw_inventories.tsv
```diff
@@ -53,3 +53,43 @@ roseslabrada2015mako.pdf		Mako		i		i
...
+						ɛ		
+				u		u	"For the high vowels, /i/ and /u/, tensing is particularly noticeable in open syllables"	
+						ʊ		
+				o		o	"For the mid vowels, /e/ and /o/, tensing is more noticeable when tone is high"	
+						ɔ		
+				ɑ		ɑ		
+				˦	high/rising	˦		
+						˩˥	"When a low or falling tone precedes a high tone, the high tone begins with an upward pitch, like a rising tone, whereas elsewhere it is a high tone."	
+				˨		˨		
+						˥˩	"where a high/rising tone precedes a low tone, the low tone begins with a downward pitch, that is, it appears to be like a falling tone, but not falling as much as a true falling tone"
```
drammock, line 94: I'd interpret the text here to imply a 3-1 sequence (mid-falling-to-low ˧˩) rather than 5-1 (high-falling-to-low)

raw-data/UW/uw_inventories.tsv
```diff
@@ -53,3 +53,43 @@ roseslabrada2015mako.pdf		Mako		i		i
...
+						ɪ		
+				e		e	"For the mid vowels, /e/ and /o/, tensing is more noticeable when tone is high"	
+						ɛ		
+				u		u	"For the high vowels, /i/ and /u/, tensing is particularly noticeable in open syllables"	
+						ʊ		
+				o		o	"For the mid vowels, /e/ and /o/, tensing is more noticeable when tone is high"	
+						ɔ		
+				ɑ		ɑ		
+				˦	high/rising	˦		
+						˩˥	"When a low or falling tone precedes a high tone, the high tone begins with an upward pitch, like a rising tone, whereas elsewhere it is a high tone."
```
drammock, line 92: similar comment as for the low falling allophone: I'd put this at 3-5 (˧˥) rather than 1-5

drammock:
> The last three commits are the only ones that make any changes. The others are there because my repository got out of sync, and they were included in the previous pull request.

This is not a problem since we "squash and merge" anyway.

raw-data/UW/uw_inventories.tsv
bnk7: On the bottom of page 13 and top of page 14 it says, "The /r/ is actually a trill, but in fast speech often becomes flap." Should I include ɾ as an allophone?

raw-data/UW/uw_inventories.tsv
bnk7: Understood.

raw-data/UW/uw_inventories.tsv
drammock: ah, I missed that. Yes, please add the flap as allophone and leave phoneme as `r`.

drammock:
thanks @bnk7!

PR #294, opened 2020-07-18
add updated Jamsay inventory

bnk7:
This is based on the same text that is already cited in the references file.


bambooforest:
@bnk7 -- i think it would be better if we simply overwrite the former inventory for this doculect here:

https://github.com/phoible/dev/blob/master/raw-data/GM/gm-afr-inventories.tsv#L10294-L10355

instead of adding a new one to UW.

bambooforest:
@bnk7 -- looks great. many thanks!

PR #296, opened 2020-07-27
regenerates data for PR #295

bambooforest:


bambooforest:
We decided to:

* remove the geminate consonants because they are reported as word medial consonant geminates, but are not phonemic in the description (page 36

* @bnk7 initially remove those the nasalized vowels, but I think they should stay because on page 41 it says that:

"The independently nasalized vowels are represented with a following superscript n."

meaning, the five vowels (minus the two above) are actual stand-alone word forms in the language, but there are no words represented by the values õː and ẽː.

* trill is still there

bambooforest:
Sorry, didn't mean to close the PR. :)

drammock:
Sorry....  I figured it out.  The changes here reflect edits to the source file made in **2 PRs**: #294 and #295. Merging.  Thanks for your patience with my absent-mindedness @bambooforest

PR #298, opened 2020-07-28
add Najamba Dogon

bnk7:


bnk7:
What is the proper notation for a nasal with an oral release? I currently have mᵇ.

raw-data/UW/uw_inventories.tsv
```diff
@@ -192,3 +192,176 @@ heath2016donnoso.pdf	donn1239	Donno So Dogon	Wendegele	pʰ	"Voiceless stops occu
...
+				uː		uː		
+						ũː	"a nasal... or nasalized sonorant... at the onset of a syllable... induces nasalization on an intervocalic oral sonorant... at the beginning of the following syllable. Phonetically... the nasalization extends to adjoining vowels"	
+				o	+ATR	o		
+						õ	"a nasal... or nasalized sonorant... at the onset of a syllable... induces nasalization on an intervocalic oral sonorant... at the beginning of the following syllable. Phonetically... the nasalization extends to adjoining vowels"	
+				oː		oː		
+						õː	"a nasal... or nasalized sonorant... at the onset of a syllable... induces nasalization on an intervocalic oral sonorant... at the beginning of the following syllable. Phonetically... the nasalization extends to adjoining vowels"	
+				ɔ	-ATR	ɔ		
+						ɔ̃	"a nasal... or nasalized sonorant... at the onset of a syllable... induces nasalization on an intervocalic oral sonorant... at the beginning of the following syllable. Phonetically... the nasalization extends to adjoining vowels"	
+				ɔː		ɔː
```
bnk7, line 361: When I go to this file, GitHub tells me that there is "illegal quoting" in line 361 (so it can't make the file "beautiful and searchable"), but I'm not sure what the issue is because there are no quotation marks in this line.

drammock: The problem is probably [line 363](https://github.com/phoible/dev/blob/0f42b73f14effcd2765b27b433057a4b853ee2aa/raw-data/UW/uw_inventories.tsv#L363) actually.  Sometimes GitHub forgets how to count when things like this happen

drammock:
> ᵇ

There is currently no usage of `ᵇ` anywhere in PHOIBLE (nor `ᵐ`, incidentally).  What you've done is one possibility but probably not consistent with what we've done before.  Other possibilities are `mb` (this is definitely already used for "prenasalized stops" though) and `mm͊` (the tilde-with-slash means "denasalized" in the extensions of the IPA used to transcribe disordered / pathological speech).  @bambooforest do you have insight / opinion here?  I suspect we've not been careful enough to distinguish these things (i.e., prestopped nasals vs postnasalized stops, prenasalized stops vs nasals with oral release).

@bnk7 just to clarify, "oral release" could also mean "while phonating with velum down, the (lip) closure is opened before phonation ends" which might indicate a transcription like `mə̃`.  Is there indication in the source which one is meant?

bnk7:
> > ᵇ
> 
> There is currently no usage of `ᵇ` anywhere in PHOIBLE (nor `ᵐ`, incidentally). What you've done is one possibility but probably not consistent with what we've done before. Other possibilities are `mb` (this is definitely already used for "prenasalized stops" though) and `mm͊` (the tilde-with-slash means "denasalized" in the extensions of the IPA used to transcribe disordered / pathological speech). @bambooforest do you have insight / opinion here? I suspect we've not been careful enough to distinguish these things (i.e., prestopped nasals vs postnasalized stops, prenasalized stops vs nasals with oral release).
> 
> @bnk7 just to clarify, "oral release" could also mean "while phonating with velum down, the (lip) closure is opened before phonation ends" which might indicate a transcription like `mə̃`. Is there indication in the source which one is meant?

This is what the document says:
"Based on impressionistic transcriptions, initial (and for that matter medial) “mb” has a range of articulations including [mᵇ], i.e. an m with a faint oral release." (page 14)
This is under the consonant clusters heading, so I interpreted /mb/ as a series of two phonemes rather than a prenasalized stop. However, if [mᵇ] is underlyingly /mb/, maybe I should take it out of the inventory altogether.

drammock:
> This is under the consonant clusters heading

Looking closer, I noticed this in your source document:  "The m in several of these words is optionally (or dialectally) pronounced with a brief oral release."

so I see now that `mᵇ` is an allophone of `m`, and it wouldn't make sense to treat it as a sequence.  This also I think rules out the `mə̃` interpretation.  So we're still left with the question of `mᵇ` versus `mb`.  `ᵇ` is also never appearing in our Allophones column, so I'd lean toward `mm͊` I guess (since `mb` is used in Phoible for "prenasalized stop"), even though it means introducing a new diacritic (at least it's one from the extIPA;  I feel like using `ᵇ` opens the door to a whole bunch of other superscripted glyphs that we don't want to start using if we can avoid it).  @bambooforest you OK with that?

bambooforest:
@bnk7 -- i like that you caught "Glottal stop /ʔ/ occurs only in the usual unh-unh type of interjections and does not have phonemic status." in the description of Najamba, even though it's listed in the chart! very detailed work. great job.

i see /r/ in the Najamba consonant chart, but i don't see anything about it in the description. did you? i suppose we should add it, unless there's some passage like the one above that i've missed.

@drammock -- as per our discussion, the use of ExtIPA above sounds good to me. can you update them, if you haven't already, @bnk7 ?

then i think we can merge.

PR #303, opened 2020-08-02
adds phono squib for source

bambooforest:
@drammock -- same as last PR, which i'll merge after i deal with the one by bnk before that.


drammock:
merge when you're ready, @bambooforest

PR #309, opened 2020-09-06
Add Yorno So and Ben Tey

bnk7:


bambooforest:
@bnk -- i see you added the other marginals, but not <f>. i wonder if we should add it because Heath says: "f does not occur in my data. It may occur for some speakers in Fulfulde and French borrowings", which i agree is bit ambiguous, but compare that we add tS.

bnk7:
> @bnk -- i see you added the other marginals, but not . i wonder if we should add it because Heath says: "f does not occur in my data. It may occur for some speakers in Fulfulde and French borrowings", which i agree is bit ambiguous, but compare that we add tS.

I didn't add ʃ either, although it appears in the inventory as marginal. Heath says: "ʃ does not occur in my data, and I did
not notice any strong tendency to phonetically palatalize s before i or other front vowels." My thoughts were to include t̠ʃ because the author does know of one example.
Let me know what you think.

bambooforest:
@bnk7 -- looks good!

Issue #314, opened 2020-10-28
Error in sources


tresoldi:
There is a small error in the Bibtex entry for `saphon`, a missing `"and"` that makes the entry invalid: on line 37 of https://github.com/phoible/dev/blob/master/data/phoible-references.bib we have `Michael, Lev, Tammy Stark, and Will Chang` instead of `Michael, Lev, *and* Tammy Stark, and Will Chang`.

It is the only error when parsing the whole the whole file. I cannot make a PR due to permissions.

bambooforest:
Thanks @tresoldi . You should be able to make a PR after you've forked phoible dev.

tresoldi:
Oh, I though you used branches, sorry. I'll make one right now.

bambooforest:
We do use branches -- but only from forks. :)

bambooforest:
Closed by #315

PR #315, opened 2020-10-28
Fix error in bibtex entry

tresoldi:
This PR fixes a tiny error, as mentioned in https://github.com/phoible/dev/issues/314


bambooforest:
@tresoldi -- thanks. @drammock - i doubled checked this with `biber --tool -V phoible-references.bib`. looks good. merging.

PR #325, opened 2020-11-22
fix Ó Sé's surname

jimregan:


bambooforest:
many thanks @jimregan

PR #357, opened 2022-12-30
fix some tests

### LINKED ISSUE #291: Update ISO codes
Changes in ISO 639-3 need to be revisited:

daf
duj
wit
kxl
krm
kxu
drh
dit
mwd
qgu
bjd
ggr
yiy

drammock:
closes #291 

this updates some ISO codes based on code mergers, and adds exceptions for others where we're using a retired code (usually due to a split) and we don't know for sure (yet) which of the split-created codes we should use for the inventory.

Should hopefully make one of the tests start passing.


drammock:
@bambooforest this is ready from my end, pending our offline discussion about manually changing vs auto-importing-from-glottolog of merged ISO codes.

drammock:
(nb: only failing test now is the "distinct feature vectors" test, which is being addressed separately. All the inventory metadata tests pass now!)

PR #359, opened 2023-01-26
Fix pharyngeal features

drammock:
This assigns `-ATR,-RTR` to all non-pharyngeal base glyphs and assigns `+RTR,-ATR` to the pharyngeal base glyphs ʕ and ħ (the pharyngealized diacritic ˤ already has those features).


drammock:
after this the unique failed contrast pairs drops from 266 to 196 🎉 

marking as draft though because I realized I haven't yet done this for the precomposed "special-cased" phonemes (the ones where we can't easily compute the features automatically)

bambooforest:
Wow! Awesome.

A note here that we should update this page after all changes (which we link to from the [faq](https://phoible.org/faq#phonological-features-in-phoible):

https://github.com/phoible/dev/tree/master/raw-data/FEATURES

drammock:
@bambooforest this PR has the requested README addition, I think?  In ef31d26. So this is ready from my end.

PR #360, opened 2023-01-29
Add the Talian and Mojeno Trinitario inventories

Alessioryan:


Alessioryan:
The Bibtex GUIs were formatting the file differently, adding watermarks for the software I was using. I ended up adding it manually to the end of the document, viewing it as a plain text file.

Alessioryan:
The Bibtex GUIs ended up adding watermarks or some type of metadata which resulted in the document having more changes than I actually added. I ended up downloading the citations manually and adding them to the bottom of the document.

data/phoible-references.bib
```diff
@@ -42457,3 +42457,23 @@ @book{heath2015togokan
 	Title = {A Grammar of Togo Kan (Dogon language family, Mali)},
 	Url = {https://deepblue.lib.umich.edu/handle/2027.42/123066},
 	Year = {2015}}
+
+@article{rose2021mojenotrinitario.pdf,
```
bambooforest, line 42461: @Alessioryan -- there shouldn't be the ".pdf" in the ID (like the rest of the entries). Please remove.

data/phoible-references.bib
```diff
@@ -42457,3 +42457,23 @@ @book{heath2015togokan
...
+	title={Mojeño Trinitario},
+	volume={52},
+	DOI={10.1017/S0025100320000365},
+	number={3},
+	journal={Journal of the International Phonetic Association},
+	publisher={Cambridge University Press},
+	author={Rose, Françoise},
+	year={2022},
+	pages={562–580}}
```
bambooforest, line 42470: @Alessioryan -- in latex / bibtex we use "--" for the page range; it becomes an en dash. Please update.

data/phoible-references.bib
```diff
@@ -42457,3 +42457,23 @@ @book{heath2015togokan
...
+
+@article{guzzo2022talian.pdf,
+	title={Brazilian Veneto (Talian)},
```
bambooforest, line 42473: @Alessioryan -- articles titles will also get automatically lower-cased, so they proper nouns like language names need to be escaped, e.g., {Brazilian Veneto (Talian)} --> {{Brazilian Veneto (Talian)}}. Please update.

data/phoible-references.bib
Alessioryan: Done! I misunderstood a comment in a previous email but both are fixed now.

data/phoible-references.bib
Alessioryan: Done!

data/phoible-references.bib
Alessioryan: Makes sense, I just updated it. Thanks!

drammock:
formatting look reasonable; @bambooforest did you already review the source / inventory ?

bambooforest:
@Alessioryan -- looks good in general, some comments below.

One tricky thing that we need to update is our idiosyncratic usage of place matching, i.e., the post-alveolar affricates, e.g., t̠ʃ, get the retraction symbol (U+0320).

Looking at Brazilian Veneto (Talian) (this is a very detailed description!), I'm not sure we should list post-alveolar fricatives (/S Z/) as marginals (not IPA getting garbled here and below). The author writes:

>>> Talian does not have post-alveolar fricatives (/S Z/), which are also absent in other Veneto varieties (Zamboni 1974, Canepari 1976), but exist in Brazilian Portuguese. In loanwords from Brazilian Portuguese, /S Z/ are typically adapted as /s z/: [so»Rasko] (for [Su»xaskU] churrasco ‘barbecue’), [si»nEla] (for [Si»nEl“] chinela ‘sandal’), [zaka»RE] (for [Zaka»RE] jacaré ‘cayman’), [»zeÉito] (for [»ZeÉitU] jeito ‘manner’). **However, Talian speakers who are dominant in Portuguese or use Portuguese on a regular basis employ target /S Z/ in such loanwords (Margotti 2004).**

Since as highlighted, it would be just those speakers that are dominant in Portuguese. @drammock thoughts?

Also, I'm confused about the diphthongs. They are listed here, but the author writes:

>>> Figure 6 shows Talian’s surface falling diphthongs.3 Two observations suggest that these diphthongs are VV sequences where the second vowel is diphthongized. First, these diph- thongs can be variably produced as hiatuses. Second, surface diphthongs resulting from the pluralization of stems ending in /l/ exhibit metaphony, which suggests that the stressed vowel and the final vowel are originally in separate syllables: /fa»zol/ fasol ‘bean.SG’→ [fa»zoÉi] ∼ [fa»zuÉi] fasoi ‘bean.PL’. Talian also seems to have the diphthong /aÉu/; however, many words with such a diphthong derive from words with hiatus: /pa»uRa/ paura ‘fear’, /spaÉu»Ra/ spaurà ‘frightened’.

>>> [3] Falling diphthongs are understood here as diphthongs whose first element is more prominent (see Crystal 2008).

But then does seem to list minimal pairs, e.g vao vs. deo.

I'd also add to /j w/ this comment:

>>> It is open to discussion whether /j w/ are phonemes in Talian.

----

Mojeño Trinitario

This inventory looks good to me. We unfortunately haven't been consistent with diphthongs with regard to our conventions -- although we have been consistent to the source materials for the most part. 

@drammock:

>>> Vowel plots have been produced with the phonR package of the R software (McCloy 2016, R Core Team 2019).

:)

drammock:
> it would be just those speakers that are dominant in Portuguese. @drammock thoughts?

"occurs only in loanwords" is usually grounds for including as marginal. I don't think we've ever confronted the situtation here: "occurs only in loanwords in the speech of bilinguals". The crucial question is probably "what fraction of the native speakers are bilingual w/ Portuguese"?  If it's well above 50% I would probably include these.  On the assumption that we don't know that number, however, I would say don't include them...  To give an analogy: many american english speakers who know modern standard chinese might instinctively insert tones into chinese loanwords, but that doesn't mean that those tonemes are marginal tonemes of English.

> Also, I'm confused about the diphthongs.

I would not include them as phonemes, based on the author's reasoning that there are multiple indicators that they are underlyingly VV sequences.

> Vowel plots have been produced with the phonR package 

Nice to see someone using its diphthong plotting abilities for once!  Although it's much nicer when there are more than 2 measurement times per vowel.

Alessioryan:
@drammock @bambooforest Sounds good, I'll add the changes in my next push.

Just to confirm:
- The Talian PDF states that "Today, virtually all speakers of Talian also speak Brazilian Portuguese". This implies that /S Z/ are not marginal in the speech of the vast majority of Talian speakers, so I'll go back and add them as independent phonemes.
- Change the affricate to t̠ʃ.
- Remove the diphthongs.
- Add the comment on /j w/.
Should I add the comment about diphthongs in the "notes" section?

drammock:
> The Talian PDF states that "Today, virtually all speakers of Talian also speak Brazilian Portuguese". This implies that /S Z/ are not marginal in the speech of the vast majority of Talian speakers, so I'll go back and add them as independent phonemes.

I still think this is the wrong call.  Looking at how the author phrased it, they start by saying:

> Talian does not have post-alveolar fricatives (/ʃ ʒ/)

and then go on to say:

> In loanwords from Brazilian Portuguese, /ʃ ʒ/ are typically adapted as /s z/

and then discuss BP-dominant speakers as the *exception*. True, they do say that virtually all Talian speakers nowadays also speak BP, but they don't say they are all *dominant* or *speak it regularly*.

bambooforest:
Agreed. Rest looks good. Never hurts to have copious notes if they don't take you much time. Thx!

Alessioryan:
Fixed the issues! I added the diphthong comment only on the vowels which diphthongize (i.e. /i/, /o/).

drammock (approved):
LGTM. @bambooforest I let you merge if you're happy

PR #375, opened 2024-07-04
Update phoible-references.bib

inigourrestarazu:
I spotted a typo in the sections of a book, West African Language Data Sheets.

It says Langauge, instead of Language. I also found it in another reference of a Bantu language. I replace all of them


bambooforest:
@inigourrestarazu -- many thanks for this. and @LinguList for sending it this way.

PR #365, opened 2023-04-07
Fix various feature bugs

drammock:
This PR does a few things:

- fix bug with special-cases feature table (lookups were happening, but not being properly handled) (2a4306c)
- when building feature vectors, clicks are now handled like modifiers (not like base glyphs), which makes it easier to get the correct features (previously they were incorrect) (a1007e7)
- special-cases a few clicks that are hard to get right automatically (aspirated ejective clicks) (fa0d362)
- stop ignoring the retraction diacritic on `t` and `d` (affricate features are unchanged, and now some australian languages have proper feature distinguishability) (8e7355a)
- generally improves the iterative feature-vector-building process, which results in a few other feature vectors changing (mostly pre-nasalized stops or affricates, they gain one or a few contour features). I consider this also a bugfix (it makes sense that, eg., `nts` would have a contour feature for [strident] or for [nasal]
- adds a `.lintr` file so my IDE isn't full of wiggly blue underlines while I code (if you don't want this I can add it to .gitignore and just keep it locally)

PR #387, opened 2026-07-08
fix: clicks no longer silently deleted when adjacent to a modifier

varunghat:
Fixes clicks getting silently deleted when reordering an adjacent modifier
(e.g. kǀʰʷ -> kʷʰʷ) in order_ipa()

Clicks were typed the same as modifiers ("M") in make_typestring()
Gave clicks their own typestring code ("K"). 

Doesn't touch feature-helper-functions.R's separate click="M" typing, so feature computation is unaffected.

### DIFF
--- a/scripts/aggregation-helper-functions.R
+++ b/scripts/aggregation-helper-functions.R
@@ -298,7 +298,14 @@ order_ipa <- function(strings, keep_stars=FALSE, keep_brackets=TRUE) {
             typestring <- paste(typstr, collapse="")
         }
         ## If a diacritic comes right after a modifier letter, swap their order
-        while (stri_detect_fixed(typestring, "MD")) {
+        if (stri_detect_fixed(typestring, "MD")) {
+          ixs <- stri_locate_all_fixed(typestring, "MD")[[1]]
+          for (row in seq_len(dim(ixs)[1])) {
+              ix <- ixs[row, 1]
+              if (string[ix] %in% clicks) next
+              # SWAP LOGIC REMAINS SAME
+          }   }
+
             ix <- stri_locate_first_fixed(typestring, "MD")[1]
             # If the modifier letter is a click, don't swap it with the diacritic
             if (string[ix] %in% clicks) break
--- a/scripts/aggregation-helper-functions.R
+++ b/scripts/aggregation-helper-functions.R
@@ -235,13 +235,13 @@ make_typestring <- function(strings, ...) {
         codepts <- get_codepoints(chars)
         codepts[codepts %in% get_codepoints(base_glyphs)] <- "B"
         codepts[codepts %in% get_codepoints(modifiers)] <- "M"
-        codepts[codepts %in% get_codepoints(clicks)] <- "M"
+        codepts[codepts %in% get_codepoints(clicks)] <- "K"
         codepts[codepts %in% get_codepoints(diacritics)] <- "D"
         codepts[codepts %in% get_codepoints(contour_glyphs)] <- "C"
         codepts[codepts %in% get_codepoints(tones)] <- "T"
         codepts[codepts %in% get_codepoints(null_phone)] <- "N"
         codepts[codepts %in% get_codepoints(disjunct)] <- "|"
-        missed <- !codepts %in% c("B", "M", "C", "D", "T", "N", "|")
+        missed <- !codepts %in% c("B", "M", "K", "C", "D", "T", "N", "|")
         if (any(missed)) {
             warning(paste("Unfamiliar glyph components.", "Phone:", string,
                           "Codepoint:", codepts[missed]),
--- a/scripts/aggregation-helper-functions.R
+++ b/scripts/aggregation-helper-functions.R
@@ -235,6 +235,7 @@ make_typestring <- function(strings, ...) {
         codepts <- get_codepoints(chars)
         codepts[codepts %in% get_codepoints(base_glyphs)] <- "B"
         codepts[codepts %in% get_codepoints(modifiers)] <- "M"
+        # Clicks are typed as "M" so they factor into feature-vector assignment the same way modifiers do.
         codepts[codepts %in% get_codepoints(clicks)] <- "M"
         codepts[codepts %in% get_codepoints(diacritics)] <- "D"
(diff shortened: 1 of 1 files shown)


scripts/aggregation-helper-functions.R
```diff
@@ -235,13 +235,13 @@ make_typestring <- function(strings, ...) {
         codepts <- get_codepoints(chars)
         codepts[codepts %in% get_codepoints(base_glyphs)] <- "B"
         codepts[codepts %in% get_codepoints(modifiers)] <- "M"
-        codepts[codepts %in% get_codepoints(clicks)] <- "M"
+        codepts[codepts %in% get_codepoints(clicks)] <- "K"
```
drammock: I checked `git blame` for this line, which led me to #365. It's PR description says (among other things):

> when building feature vectors, clicks are now handled like modifiers (not like base glyphs), which makes it easier to get the correct features (previously they were incorrect)

SO I think this change will likely result in incorrect feature vectors for clicks. I'd rather try changing the line in `order_ipa` that checks for `modifiers %in% mods` (see other comment)

scripts/aggregation-helper-functions.R
drammock: I think this is the correct fix

```suggestion
                string[span] <- c(clicks, modifiers)[c(clicks, modifiers) %in% mods]
```

drammock (changes requested):

scripts/aggregation-helper-functions.R
varunghat, line 309: We need to add an if statement here to prevent the swap applied on clicks typed as "M" 
```r
if (string[ix] %in% clicks) break
```

scripts/aggregation-helper-functions.R
varunghat: I checked the diff between output files generated with the original code, previous fix (clicks as "K") and this suggested fix. There's a few things to note: 
1. The feature vectors do change, but worth noting the original (unfixed) code does not produce a valid phoible.csv file that matches the one in the repo. Running a fresh install crashes in add-features.R with the following error:
`GlyphID 0353 (frictionalized) used on base glyph k was not handled.` 

This probably means the current phoible.csv was obtained by running add-features.R on a pre-existing phoible-nofeats.csv?

2. The original code and the suggested fix change the ordering of some modifiers. For example: in kǃ̠ʰ (Present in HAZDA and SANDAWE), the original code changes it to k̠ʰʰ, then crashes with the above error. The suggested fix moves the diacritic onto the base k instead (k̠ǃʰ). This implies the feature vectors will be different but I'm not sure if this is intended behaviour. 

This is because the above fix only solves for "MM" (two modifiers appearing sequentially) and not for "MD" (which is the above example - "BMDM")
So we need to further add in line 302 under ```while (stri_detect_fixed(typestring, "MD"))```

```r
            if (string[ix] %in% clicks) break
```

(Please check the comment below)

3. Finally, I compared the files generated with the previous fix (clicks as "K"), this complete fix and the  phoible.csv currently committed in the repo: All three are identical.

scripts/aggregation-helper-functions.R
```diff
@@ -299,6 +299,7 @@ order_ipa <- function(strings, keep_stars=FALSE, keep_brackets=TRUE) {
         ## If a diacritic comes right after a modifier letter, swap their order
         while (stri_detect_fixed(typestring, "MD")) {
```
varunghat: Note: This simple fix works for all the current data, BUT could cause an issue if there's a phoneme with click-diacritic pair AND an unrelated modifier-diacritic pair in the same string. For that, we will need to locate all "MD" with this:

```suggestion
        if (stri_detect_fixed(typestring, "MD")) {
          ixs <- stri_locate_all_fixed(typestring, "MD")[[1]]
          for (row in seq_len(dim(ixs)[1])) {
              ix <- ixs[row, 1]
              if (string[ix] %in% clicks) next
              # SWAP LOGIC REMAINS SAME
          }   }

```

drammock: the case of `kǃ̠ʰ` shows that we do indeed need some extra logic like this; the "combining minus below" should stay on the `ǃ` and not end up on the `k`.

scripts/aggregation-helper-functions.R
```diff
@@ -312,7 +313,7 @@ order_ipa <- function(strings, keep_stars=FALSE, keep_brackets=TRUE) {
             for (row in seq_len(dim(ixs)[1])) {
                 span <- ixs[row,1]:ixs[row,2]
                 mods <- string[span]
-                string[span] <- modifiers[modifiers %in% mods]
+                string[span] <- c(clicks, modifiers)[c(clicks, modifiers) %in% mods]
```
drammock, line 326: so should this change be reverted in light of your new suggested logic?

drammock (commented):
@varunghat could you push whatever changes are needed get the code into the state that you think is easiest to understand/maintain that still preserves parity with the current state of the output file (phoible.csv)?  Then ping me and I'll look again / merge.

scripts/aggregation-helper-functions.R
varunghat: The current change (your suggestion+the extra line I added) still works for kǃ̠ʰ ("BMDM"), the case where it wouldn't won't work is a case like "*MD*MD*", this pattern does not occur anywhere in our current data (including the new upcoming data)

scripts/aggregation-helper-functions.R
varunghat: No need to revert, new suggested logic (latest commit) is based on the above suggestion, just added a line accounting for a separate edge case.

varunghat:
@drammock Latest commit on this PR has the change which is probably easiest to understand/maintain, with minimal code change. Added some comments to explain this edge case. Still preserves parity with the current state of the output file phoible.csv .

drammock (approved):

varunghat:
@drammock The "suggested" code change was more of a comment (while loop -> if), not meant to be a syntactically correct drop-in replacement. My bad for posting it as a suggestion block instead of a plain code block. 

Current code is now syntactically incorrect and throws an error. Created a hotfix #390 with correctly formatted code, verified with regenerating phoible.csv, still preserves parity with currently committed phoible.csv

Issue #399, opened 2026-08-19
Update InventoryID to non-integer strings and add soft delete


varunghat:
Update InventoryID from ordered integers to hex strings to remove linearity 

- Ensure all are non-integer so they throw a warning instead of silently accepting an integer ID during access
- Soft delete languages that get removed, so that InventoryIDs stay consistent between releases
- Add mapping file from old ID to new ID and entry in Readme
- Add a test to ensure IDs are consistent with changes
- Website links need to redirect from old view/InventoryID pages to new view/InventoryID pages, with an empty page for deleted languages (Might need to open this issue separately on the cldf side?)

PR #405, opened 2026-08-20
More tests, workflows PR comment for feature test

varunghat:
Added the following regression tests:
- Clicks and modifiers get typed as "M", which caused some issues previously with clicks and diacritics getting swapped where they shouldn't. Added a test to make sure the behaviour is as intended.
- c-cedilla is a special case handled in denorm(), as well as 20 special-cases for features. Added a test to ensure these are not changed

Added the following new test
- test-cross-inventory-integrity.R checks if 
  - any (Source, Glottocode, LanguageName, SpecificDialect, BibtexKey) repeats with the same InventoryID , which implies there's two languages with the same InventoryID 
  - Checks if all BibtexKey map to a valid Bibtex entry in phoible-references.bib
  - Checks if each InventoryID in bibtex mapping file is present in the final phoible.csv and vice-versa
  
Added the following extra test to test-inventory-metadata.R
- Check if the ISOcodes present in glottolog match the ISOcodes present in phoible (With certain exceptions where glottolog keeps the older ISOcodes)

Excluded test-phonemes.R from github workflows test suite, but added it as a separate job which writes the output as a comment in the PR 

Added tasks to pixi.toml (exactly the same as #398)


varunghat:
Split into two workflows now (Since PR from forked repo does not have write permissions) 

First workflow runs the Feature test (test-phonemes.R) first and stores data. Then runs the tests suite. 
Second workflow runs on completion of the first, and writes the PR comment. 

Verified the workflow on my forked branch, PR comment works with PR from a forked repo. Will work once the new workflow file gets added to master. 

Current test fails are the 5 ISOcodes that do not match with glottolog (More info in this PR that fixes them: #403)

.github/workflows/main.yml
```diff
@@ -5,17 +5,42 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
-
+
```
drammock:
```suggestion

```
stray whitespace.

.github/workflows/main.yml
```diff
@@ -5,17 +5,42 @@ on:
...
   pull_request:
     branches: [ master ]
-
+ 
 jobs:
   test-data:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      actions: write
```
drammock: why does this job need `actions: write`?

.github/workflows/main.yml
```diff
@@ -5,17 +5,42 @@ on:
...
+          path: |
+            tests/phoneme_feature_test_results.txt
+            tests/pr_number.txt
+
       - name: run tests
         working-directory: tests
         run: |
-          Rscript -e "testthat::test_dir('.', stop_on_failure=TRUE)"
+            Rscript -e "testthat::test_dir('.', filter = 'phonemes', invert = TRUE, stop_on_failure = TRUE)"
+
```
drammock:
```suggestion

```
stray whitespace

tests/test-cross-inventory-integrity.R
```diff
@@ -0,0 +1,100 @@
...
+##      phoible-references.bib.
+##   3. Every InventoryID in the mapping files exists in the aggregated
+##      data, and vice versa.
+
+library(dplyr, warn.conflicts=FALSE)
+library(testthat, warn.conflicts=FALSE)
+
+## load PHOIBLE data
+phoible_data_file <- file.path("..", "data", "phoible.csv")
+phoible_col_types <- readr::cols(InventoryID="c", Marginal="l", .default="c")
```
drammock, line 15: reading `InventoryID` as character?  is this a mistake, or anticipatory of #399?

tests/test-regression-click-modifier.R
```diff
@@ -0,0 +1,81 @@
...
+    result <- order_ipa("ǀǁʰ")
+    expect(stri_detect_fixed(result, "ǀ"),
+           paste("First click character was lost from", shQuote(result)))
+    expect(stri_detect_fixed(result, "ǁ"),
+           paste("Second click character was lost from", shQuote(result)))
+    expect(nchar(result) == 3,
+           paste("Expected a 3-character result, got", shQuote(result),
+                 "(", nchar(result), "characters)"))
+})
+
```
drammock:
```suggestion
```
no point to test this as you can't have 2 click glyphs in a row (click glyphs must always be accompanied by a preceding base glyph that denotes the place of back closure; the click glyph denotes front closure). In other words ǀǁʰ would really have to be kǀkǁʰ or qǀkǁʰ or something like that

tests/test-regression-click-modifier.R
```diff
@@ -0,0 +1,81 @@
...
+    result <- order_ipa("ǀ̥ǁ̥")
+    expect(stri_detect_fixed(result, "ǀ"),
+           paste("First click character was lost from", shQuote(result)))
+    expect(stri_detect_fixed(result, "ǁ"),
+           paste("Second click character was lost from", shQuote(result)))
+    expect(nchar(result) == 4,
+           paste("Expected a 4-character result, got", shQuote(result),
+                 "(", nchar(result), "characters)"))
+})
+
```
drammock: I think we can skip this one too, for the same reason as above
```suggestion
```

tests/test-regression-special-casess.R
```diff
@@ -0,0 +1,33 @@
...
+                       "qχ", "bβ", "bv", "dð", "dz", "dʒ", "ɖʐ", "ɟʝ", "ɡɣ",
+                       "ɢʁ", "kp", "ɡb")
+    special_feats <- read.csv(file.path("..", "raw-data", "FEATURES",
+                              "special-feature-table.csv"),
+                              fileEncoding="UTF-8")
+    missing <- setdiff(special_cases, special_feats$segment)
+    expect(length(missing) == 0,
+           paste("special_cases pair(s) missing from special-feature-table.csv:",
+                 paste(missing, collapse=", ")))
+})
```
drammock: this is OK, but it seems like we could achieve the same thing by calling `stopifnot(nrow(base_vec) == 1)` right after the lookup, to make sure we got back exactly one row:

https://github.com/phoible/dev/blob/6da1f9ec7c7f941ce647a0d9749a8a525901fb5c/scripts/add-features.R#L145-L146

drammock (changes requested):

tests/test-cross-inventory-integrity.R
varunghat: Anticipatory (can change if needed) and also because I'm reading bibtex mapping file as all characters:

https://github.com/varunghat/Phoible-dev/blob/feat/more-tests/tests/test-cross-inventory-integrity.R#L18-L19

Can also change if needed to have consistency with rest of tests (read as integer) until #399

drammock: I suppose it doesn't matter much since all you're doing is merging tables on that key, so OK as-is.

.github/workflows/main.yml
varunghat: Removed whitespace in https://github.com/phoible/dev/pull/405/commits/520b59211adbbda89351c2ae076a91adedc4ea28

.github/workflows/main.yml
varunghat: Job does not need actions: write. I wrongly assumed writing file needed it. Removed in https://github.com/phoible/dev/pull/405/commits/520b59211adbbda89351c2ae076a91adedc4ea28. Tested on separate fork.

.github/workflows/main.yml
varunghat: Removed in https://github.com/phoible/dev/pull/405/commits/520b59211adbbda89351c2ae076a91adedc4ea28

tests/test-regression-click-modifier.R
varunghat: Removed test https://github.com/phoible/dev/pull/405/commits/9af55311dba594647b11e4ed98436bd2b0a5b7bc

tests/test-regression-click-modifier.R
varunghat: Removed test https://github.com/phoible/dev/pull/405/commits/9af55311dba594647b11e4ed98436bd2b0a5b7bc

tests/test-regression-special-casess.R
varunghat: I agree, that is better. Avoids issue of hardcoded special_cases list in test. 

Removed the test in https://github.com/phoible/dev/pull/405/commits/f1c78ca125a6ea5cc9556e9d15b6745051391213
And renamed file to test-regression-c-cedilla.R since it only tests for c-cedilla now https://github.com/phoible/dev/pull/405/commits/792ea5a5508e5ed0bdc163a1d06610d9ec60a2ec

Added the stopifnow assert in https://github.com/phoible/dev/pull/405/commits/9f8864d53d48ae839561c0366ab6020921e16d42

Reran pipeline, phoible.csv has no diff with committed version.

drammock (approved):

